PR #6, opened 2018-08-15
Add type of the object to the index file of che-plugin-registry

skabashnyuk:
### What does this PR do?

Add type of the object to the index file of che-plugin-registry

### Reference to
https://github.com/eclipse/che/issues/10781

### DIFF
--- a/index.sh
+++ b/index.sh
@@ -7,12 +7,20 @@ echo "["
 ## now loop through meta files
 for i in "${arr[@]}"
 do
-    IFS='/' segments=($i)
     if [ "$FIRST_LINE" = true ] ; then
-        echo \{\"name\":\"${segments[2]}\",\"version\":\"${segments[3]}\"\}
+        echo "{"
         FIRST_LINE=false
     else
-        echo \,\{\"name\":\"${segments[2]}\",\"version\":\"${segments[3]}\"\}
+        echo ",{"
     fi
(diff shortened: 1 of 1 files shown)


index.sh
```diff
@@ -7,12 +7,20 @@ echo "["
...
+        echo ",{"
     fi
+    # 1. read meta.yaml to stio
+    # 2. filter lines with name,version,type
+    # 3. Replace ` :` with `":"`
+    # 4. Append `",` to the end of each line
+    # 5. Append `"` at the begining of each line
+    # 6. Remove all new lines
+    # 7. Remove last ','
+    cat $i|grep -e name -e version -e type |   sed 's/: /\":"/g'  | sed 's/$/\",/g' | sed 's/^/\"/g'  | tr -d '\n' |  sed 's/,$//g'
```
sleshchenko: Helpful comments :+1: 
I think it can be made even better by reformatting in the following way:
```bash
    cat $i | \                             # read meta.yaml to stio
       grep -e name -e version -e type | \ # filter lines with name,version,type
       sed 's/: /\":"/g'  | \              # Replace ` :` with `":"`
     ...
       sed 's/,$//g'                       # Remove last ','
```
WDYT?

garagatyi:
Forgot to add commas management to my snippet.

PR #10, opened 2018-08-19
Add field 'id' to index

garagatyi:
### What does this PR do?
Field 'id' is needed because a client should add it along with
the version to Che workspace attributes.

### DIFF
--- a/index.sh
+++ b/index.sh
@@ -17,7 +17,7 @@ do
     # 1. read meta.yaml to stio
     cat $i| \
         # 2. filter lines with name,version,type
-        grep -e name -e version -e type -e 'id:' |\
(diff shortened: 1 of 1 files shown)


index.sh
```diff
@@ -17,7 +17,7 @@ do
     # 1. read meta.yaml to stio
     cat $i| \
         # 2. filter lines with name,version,type
-        grep -e name -e version -e type |\
+        grep -e name -e version -e type -e 'id:' |\
```
skabashnyuk: LgtM
Maybe  
``` grep -e `name:` -e `version:` -e `type:` -e 'id:' |\```
wdyt

garagatyi: ok

garagatyi: fixed

PR #13, opened 2018-08-23
Add script to check plugins during the CI build 

mkuznyetsov:
### What does this PR do?

Add bash script to travis.yaml that will check that all plugin directories names align with definitions in plugins meta.yaml:
- id
- version

https://github.com/eclipse/che/issues/10839


check_plugins_location.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "!!!   version directory name: '${VERSION_DIR_NAME}' "
+    FOUND=true
+  fi
+
+  cd ../..
+done
+
+if [[ $FOUND ]];then
+  exit 1
+fi
```
skabashnyuk: new line.

mkuznyetsov:
Reworked to use `yq` for YAML parsing

garagatyi:
@mkuznyetsov @skabashnyuk I think that with the approach used in the PR testing your own contribution is a bit tricky since validation is done only on travis, but not in docker build process. 
I would consider adding it to the build process. But the current approach is better than no validation at all in any case. 
So +1 to the PR

PR #16, opened 2018-08-30
Added route and parameters to OpenShift template. Updated docs.

skabashnyuk:
### What does this PR do?
Added route and parameters to OpenShift template. Updated docs.

Reference https://github.com/eclipse/che/issues/10834

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,23 @@
 
 Execute
 ```shell
-docker build --no-cache -t eclipse/che-plugin-marketplace .
+docker build --no-cache -t eclipse/che-plugin-registry .
 ```
 Where `--no-cache` is needed to prevent usage of cached layers with plugin registry files.
 Useful when you change plugin metadata files and rebuild the image.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,14 +4,23 @@
...
 
 ## https://hub.docker.com/
 
-```eclipse/che-plugin-marketplace:latest``` image would be rebuilt after each commit in master
+```eclipse/che-plugin-registry:latest``` image would be rebuilt after each commit in master
+
+## OpenShift
+You can deploy Che plugin registry on Openshift with command.
+```
+  oc new-app -f openshift/che-plugin-registry.yml \
```
sleshchenko: As far as I understand there are default values for the specified parameters IMAGE, IMAGE_TAG, PULL_POLICY. Would it better to have the simplest command here like oc new-app -f openshift/che-plugin-registry.yml and separately describe parameters that can be overridden?

Issue #26, opened 2018-09-11
Avoid extra /*.json to API calls / more REST friendly


benoitf:
Today to get list of plug-ins, the remote call is
`http://localhost:8080/index.json`

it would be nice to cut the `/index.json` path to be more REST friendly

`http://localhost:8080/plugins`  would list all the plug-ins

Also http://localhost:8080/ could display the README.

And to grab data for a specific plug-in:
today we have:
```
http://localhost:8080/plugins/org.eclipse.che.editor.theia/1.0.0/meta.yaml
```

it would be nice to have
```
http://localhost:8080/plugins/org.eclipse.che.editor.theia/1.0.0 ?
```

```(<plugin-id>/<plugin-version>)```

and call to 
```
http://localhost:8080/plugins/org.eclipse.che.editor.theia
```
would list all versions

garagatyi:
Already fixed.

PR #27, opened 2018-09-12
Added Apache index files configuration

skabashnyuk:
### What does this PR do?
Reference issue https://github.com/eclipse/che-plugin-registry/issues/26

1. Return readme on http://localhost:8080/ 

I didn't want to maintain two readme for now and it's quite hard to add md->html convertor for this image ```FROM quay.io/openshiftio/rhel-base-httpd:latest``` that is why it shows raw md.

2. Return index.json on  http://localhost:8080/plugins

3. Return meta.yaml for plugins requests

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,1 +1,2 @@
 Header set Access-Control-Allow-Origin "*"
+DirectoryIndex index.json
--- a/.htaccess
+++ b/.htaccess
@@ -1,2 +1,2 @@
 Header set Access-Control-Allow-Origin "*"
-DirectoryIndex index.json
+DirectoryIndex README.md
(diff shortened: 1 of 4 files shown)


skabashnyuk:
old request format is working too.

benoitf:
I would say that http://localhost:8080  should provide README file
while http://localhost:8080/plugins/ should provide list of plug-ins (json)

skabashnyuk:
@benoitf I see two issues.
1.  http://localhost:8080 should provide README file - will return pure md, not html form
2. http://localhost:8080/plugins/ should provide list of plug-ins (json) - it will conflicts with meta.yaml directive.

skabashnyuk:
Content updated @benoitf @garagatyi  please take a look

benoitf:
@skabashnyuk  it seems screenshot for 3 is same than 2

skabashnyuk:
@benoitf fixed

garagatyi:
@skabashnyuk If I understand correctly index.json would not be available from the root directory after this PR and it would brake UD view that adds plugins addition, so it make sense to notify @ashumilova

skabashnyuk:
@garagatyi afaik she is aware.

PR #45, opened 2018-10-19
Update name to be Che Samples: REST API Sidecar Plugin

benoitf:
### What does this PR do?

Update name to be Che Samples: REST API Sidecar Plugin


benoitf:
@skabashnyuk to be deployed on openshiftio plugin-registry, it will be automatic ?

PR #52, opened 2018-11-02
Adding plugins/org.eclipse.che.editor.dirigible/1.0.0/meta.yaml

delchev:
…ins/org.eclipse.che.editor.dirigible/1.0.0/meta.yaml

### What does this PR do?

Adding Eclipse Dirigible Che Plugin


benoitf:
hello @delchev , as you used github to provide the change, there is no Signed-off then IP validation check is not correct

PR #61, opened 2018-11-07
Plugin for Eclipse IDE (over Broadway) as editor in Che

mickaelistria:
eclipse-ide

Signed-off-by: Mickael Istria <mistria@redhat.com>

### DIFF
--- a/plugins/org.eclipse.che.editor.eclipseide/0.0.1/meta.yaml
+++ b/plugins/org.eclipse.che.editor.eclipseide/0.0.1/meta.yaml
@@ -1,0 +1,8 @@
+id: org.eclipse.che.editor.eclipseide
+version: 0.0.1
+type: Che Editor
+name: eclipse-ide
+title: Eclipse IDE (in browser using Broadway) as editor for Eclipse Che
+description: Eclipse IDE
+icon: https://github.com/eclipse/eclipse.platform/raw/master/platform/org.eclipse.sdk/eclipse256.png
+url: https://github.com/mickaelistria/che-editor-eclipseide/releases/download/untagged-267df4c0cbc3643d425b/che-editor-plugin.tar.gz
--- a/plugins/org.eclipse.che.editor.eclipseide/0.0.1/meta.yaml
+++ b/plugins/org.eclipse.che.editor.eclipseide/0.0.1/meta.yaml
@@ -5,4 +5,4 @@ name: eclipse-ide
 title: Eclipse IDE (in browser using Broadway) as editor for Eclipse Che
 description: Eclipse IDE
 icon: https://github.com/eclipse/eclipse.platform/raw/master/platform/org.eclipse.sdk/eclipse256.png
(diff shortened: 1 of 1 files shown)


skabashnyuk:
how I can test the original image?
```
docker run -p 80:80  sopotc/eclipse-broadway
```

slemeur:
@mickaelistria : I'm eager to see that :) !!

benoitf:
I'm discussing with Mickaël, will handle it and add screencast

plugins/org.eclipse.che.editor.eclipseide/0.0.1/meta.yaml
```diff
@@ -0,0 +1,8 @@
+id: org.eclipse.che.editor.eclipseide
+version: 0.0.1
+type: Che Editor
+name: eclipse-ide
+title: Eclipse IDE (in browser using Broadway) as editor for Eclipse Che
+description: Eclipse IDE
+icon: https://github.com/eclipse/eclipse.platform/raw/master/platform/org.eclipse.sdk/eclipse256.png
+url: https://github.com/mickaelistria/che-editor-eclipseide/releases/download/untagged-267df4c0cbc3643d425b/che-editor-plugin.tar.gz
```
benoitf: Hello, here is a proposed change
```suggestion
url: https://github.com/ws-skeleton/che-editor-eclipseide/releases/download/0.0.1/che-editor-plugin.tar.gz
```

Docker Image is updated to match openshift.io security (root access not allowed) (port number is also changed to be > 1024)
Also update Eclipse .tgz downloaded, squash layers to have smaller image than the original one

Update as well the meta.yaml to match new port number and avoid secured layer (else websocket of broadway is not working with this extra layer)

benoitf:
For ppl wanted to try, here the plug-in attribute to use temporary
```
 "editor": "https://raw.githubusercontent.com/benoitf/custom-repo/master/plugins/org.eclipse.che.editor.eclipseide:0.0.1",
```

or here the full workspace.json https://gist.github.com/benoitf/4fda039549caa1374447c5dcfc2ab7bb

I've updated description of issue with a screencast

PS: as probe is missing, need to refresh the page when landing into Eclipse-IDE

plugins/org.eclipse.che.editor.eclipseide/0.0.1/meta.yaml
mickaelistria: ok!

mickaelistria:
> I've updated description of issue with a screencast

Thanks a lot!

> PS: as probe is missing, need to refresh the page when landing into Eclipse-IDE

This is the kind of thing that we may try to fix later if it becomes useful, but at the moment, I suggest we apply the changes you recommend, create a small demo of it for interested parties to consider and leave it as it until there are requests.

mickaelistria:
> are we good to merge?

I personally have no objection at all as I barely understand what I'm doing. I let you guys decide ;)

skabashnyuk:
@benoitf WDYT?

PR #64, opened 2018-12-06
Added an Helm Chart

l0rd:
### What does this PR do?

Currently the README.md had only instructions to deploy the registry on Docker and OpenShift. This PR adds an Helm Chart and the instructions to use it to deploy the registry on Kubernetes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ You can deploy Che plugin registry on Openshift with command.
 
 ## Kubernetes
 
-You can deploy Che plugin registry on Kubernetes using [helm](https://docs.helm.sh/).
+You can deploy Che plugin registry on Kubernetes using [helm](https://docs.helm.sh/). For example if you want to deploy it in the namespace `kube-che` and you are using `minikube` you can use the following command.
 
 ```bash
 
--- a/README.md
+++ b/README.md
@@ -21,6 +21,31 @@ You can deploy Che plugin registry on Openshift with command.
              -p IMAGE_TAG="latest" \
              -p PULL_POLICY="IfNotPresent"
 ```
+
+## Kubernetes
+
+You can deploy Che plugin registry on Kubernetes using [helm](https://docs.helm.sh/).
+
+```bash
+
+NAMESPACE="kube-che"
+DOMAIN="$(minikube ip).nip.io"
+helm upgrade --install che-plugin-registry \
+    --debug \
+    --namespace ${NAMESPACE} \
+    --set global.ingressDomain=${DOMAIN} \
+    ./kubernetes/che-plugin-registry/
+
+```
+
+You can use the following command to uninstall it.
+
+```bash
+
+helm delete --purge che-plugin-registry
+
+```
+
 ## Docker
 ```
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -21,6 +21,31 @@ You can deploy Che plugin registry on Openshift with command.
...
 ```
+
+## Kubernetes
+
+You can deploy Che plugin registry on Kubernetes using [helm](https://docs.helm.sh/).
+
+```bash
+
+NAMESPACE="kube-che"
+DOMAIN="$(minikube ip).nip.io"
```
benoitf: here it looks like more for minikube than k8s

kubernetes/che-plugin-registry/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+#
+# Copyright (c) 2012-2017 Red Hat, Inc.
```
benoitf: invalid copyright date

kubernetes/che-plugin-registry/templates/ingress.yaml
```diff
@@ -0,0 +1,24 @@
+#
+# Copyright (c) 2012-2017 Red Hat, Inc.
```
benoitf: Date should be 2018 (btw in 2012 it was not also started)

kubernetes/che-plugin-registry/values.yaml
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright (c) 2012-2017 Red Hat, Inc.
```
benoitf: date

kubernetes/che-plugin-registry/values.yaml
```diff
@@ -0,0 +1,15 @@
...
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+chePluginRegistryImage: eclipse/che-plugin-registry
+chePluginRegistryImagePullPolicy: Always
+chePluginRegistryMemoryLimit: 256Mi
+
+global:
+  ingressDomain: 192.168.99.100.nip.io
```
benoitf: IP is hardcoded ?

(Missing also a newline)

README.md
l0rd: I have updated the sentence above to make it clear that in the example below the target platform is minikube.

kubernetes/che-plugin-registry/Chart.yaml
l0rd: fixed

kubernetes/che-plugin-registry/templates/ingress.yaml
l0rd: fixed

kubernetes/che-plugin-registry/values.yaml
l0rd: fixed

kubernetes/che-plugin-registry/values.yaml
l0rd: Added the new line. For the ingressDomain default value I am using the hardcoded IP we use in Che Server Helm chart. We still haven't found a better solution.

PR #65, opened 2018-12-10
Use svg type icons in plugins metadata and add validation for them

mshaposhnik:
### What does this PR do?
 - Changes all existing plugins icons to svg files;
 - Add build-time validation of icons;

https://github.com/eclipse/che/issues/12095


check_plugins_images.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
```
benoitf: missing some copyright headers ?

mshaposhnik: Added in all  scripts

PR #69, opened 2018-12-17
Introduce Che/Theia dev plug-in

benoitf:
### What does this PR do?

This plug-in brings the theia development container used to build che/theia with required tools

So, create Che7 stack, pickup this plug-in and then a theia-dev container is brought 
this one: https://github.com/eclipse/che/tree/master/dockerfiles/theia-dev

It allows to work with the custom Che/Theia

Change-Id: I165d8934d82da045f2ababfc62275408d831bd50

### DIFF
--- a/plugins/org.eclipse.che.theia.dev/0.0.1/meta.yaml
+++ b/plugins/org.eclipse.che.theia.dev/0.0.1/meta.yaml
@@ -1,0 +1,8 @@
+id: org.eclipse.che.theia.dev
+version: 0.0.1
+type: Che Plugin
+name: Che Theia Dev Plugin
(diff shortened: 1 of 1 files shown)


l0rd:
@benoitf where will the hosted theia run? In the theia-dev container or in the same container where the original theia is running?

benoitf:
@l0rd it can be run on both. Probably in watch mode, dev flow will be to run it on theia-dev but to ensure it's working it will be on production one.

benoitf:
@l0rd Also it requires https://github.com/ws-skeleton/che-editor-theia/pull/11 (to decrease memory from runtime container)

PR #74, opened 2019-01-04
Add VS Code node debug and typescript plugins

tolusha:
Signed-off-by: Anatoliy Bazko <abazko@redhat.com>

### Reference issue
https://github.com/theia-ide/theia/issues/3766

### What does this PR do?
This PR adds VS Code node debug and typescript plugins into plugin registry


garagatyi:
Added WIP as it depends on non-merged broker

plugins/ms-vscode.node-debug2/1.31.1/meta.yaml
```diff
@@ -0,0 +1,10 @@
+id: ms-vscode.node-debug2
```
garagatyi: Can you elaborate why we need 2 instances of this plugin?

tolusha: They are two different node debug plugins
* ms-vscode.node-debug2 for node version >= 8.0 
* ms-vscode.node-debug2 for node version < 8.0

garagatyi: Could you make titles differ?

plugins/ms-vscode.node-debug2/1.31.1/meta.yaml
```diff
@@ -0,0 +1,10 @@
...
+version: 1.31.1
+type: VS Code extension
+name: Node Debug
+title: Node.js debugging support
+description: Node Debug is the debugger for Node.js versions >= 8.0 https://marketplace.visualstudio.com/items?itemName=ms-vscode.node-debug2
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+attributes:
+  extension: "vscode:extension/ms-vscode.node-debug2"
+  container-image: "abazko/ms-vscode.node-debug2"
```
garagatyi: Can you move images under eclipse foundation account?

plugins/ms-vscode.node-debug2/1.31.1/meta.yaml
tsmaeder: The titles are different already. What would you propose?

garagatyi: I believe they are still the same: Node.js debugging support

tolusha:
@garagatyi @sleshchenko @benoitf 
PR is ready to review again.

sleshchenko:
@tolusha Should WIP be removed from PR title now?

plugins/ms-vscode.node-debug/1.0.0/meta.yaml
```diff
@@ -0,0 +1,10 @@
+id: ms-vscode.node-debug
```
sleshchenko: Can it be `ms-vscode.node-debug-legacy` and `ms-vscode.node-debug2` just `ms-vscode.node-debug`? Because `ms-vscode.node-debug` and `ms-vscode.node-debug` are confusing names.

sleshchenko: Now I see that it is name that comes from vscode market place, right?

sleshchenko: I guess CI will fail. See required fields list https://github.com/eclipse/che-plugin-registry/blob/master/check_plugins_viewer_mandatory_fields.sh#L12

sleshchenko: @mkuznyetsov Should `firstPublicationDate` and `latestUpdateDate` would be provided automatically or it should be done manually?

It would be nice to see description for `meta.yaml` in REAME.md

plugins/ms-vscode.typescript/1.0.0/meta.yaml
```diff
@@ -0,0 +1,13 @@
+id: ms-vscode.typescript
+version: 1.0.0
+type: VS Code extension
+name: Typescript
+title: Typescript language features
+description: Typescript language features
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+url: https://github.com/tolusha/ms-code.typescript/releases/download/1.0.0/che-typescript-language.vsix
```
sleshchenko: Do we have any plan to move this plugin from `tolusha` to `che-incubator`?

tolusha: I don't mind. Should we?

sleshchenko: @garagatyi @l0rd WDYT?

sleshchenko:
screenshot_20190215_154448

Note The Travis CI build failed. possible values for category see here https://github.com/eclipse/che-plugin-registry/blob/master/check_plugins_viewer_mandatory_fields.sh#L13

plugins/ms-vscode.typescript/1.0.0/meta.yaml
garagatyi: Yes! We should not keep this in our personal repos

garagatyi: But this should not block this PR in any case

plugins/ms-vscode.node-debug/1.0.0/meta.yaml
```diff
@@ -0,0 +1,13 @@
...
+version: 1.0.0
+type: VS Code extension
+name: Node Debug (legacy)
+title: Node.js debugging support (legasy)
```
sleshchenko: typo in word `legasy`

tolusha: @sleshchenko 
You can find every tiny bugs ;)
That's really cool.

sleshchenko: :eyes: :smile_cat:

plugins/ms-vscode.node-debug/1.0.0/meta.yaml
```diff
@@ -0,0 +1,13 @@
+id: ms-vscode.node-debug
+version: 1.0.0
```
benoitf: is the version accurate ?

Also how do we ensure that the version of ""vscode:extension/ms-vscode.node-debug" is always the same (and not being upgraded automatically).
For example it may work now but not after a new release of that vscode extension

plugins/ms-vscode.node-debug/1.0.0/meta.yaml
```diff
@@ -0,0 +1,13 @@
...
+type: VS Code extension
+name: Node Debug (legacy)
+title: Node.js debugging support (legacy)
+description: Node debug (legacy) is the debugger for Node.js versions < 8.0. Together with Node Debug forms the Node.js debugging experience. https://marketplace.visualstudio.com/items?itemName=ms-vscode.node-debug
```
benoitf: quick question: for now the extensions are shipped with eclipse/che-theia-endpoint-runtime which is a "nodejs 8" version.
So how can we use that debugger that targets nodejs < 8.0 ?

benoitf: I mean, as we're always using nodejs >= 8.0 shouldn't we have a single plug-in referencing the two extensions ?

plugins/ms-vscode.node-debug/1.0.0/meta.yaml
tolusha: That's a good question. 
Initially I set the version of the extension, but Microsoft upgrades the extension automatically and after a while we will have inconsistency.
That's why I've decided to set `1.0.0` as a version.

benoitf: can we stick to a given version of microsoft extension ?

plugins/ms-vscode.node-debug/1.0.0/meta.yaml
tolusha: I am not sure If I understand you correctly. 

> shouldn't we have a single plug-in referencing the two extensions ?

I don't know if it is possible to use a single container for several plugins. 

> So how can we use that debugger that targets nodejs < 8.0 ?

No way. But user has to enable both plugins. Legacy debugger (node debug) contains functionality that is needed for another one (node debug2).

benoitf: I was thinking about UX
if I want to enable nodejs debugging, I would prefer to only enable one plug-in to my che workspace that will bring two extensions.
And if I select only one of the two it will fail ?


Also about memory if these two related plug-ins are deployed on two sidecars it's using memory

plugins/ms-vscode.node-debug/1.0.0/meta.yaml
tolusha: Yes, that's possible If use direct url to the extension [1]

[1] https://marketplace.visualstudio.com/_apis/public/gallery/publishers/ms-vscode/vsextensions/node-debug2/1.31.6/vspackage

plugins/ms-vscode.node-debug/1.0.0/meta.yaml
tolusha: Completely agree. It is not obvious that user has to enable both plugins. 

> I would prefer to only enable one plug-in to my che workspace that will bring two extensions.

Do you know how to achieve that? From my understanding we need to write a  plugin that would download both extensions and activate them, right?

plugins/ms-vscode.node-debug/1.0.0/meta.yaml
mkuznyetsov: no need to provide them manually in YAML, as they will be generated during Plugin's Registry dockerimage build

plugins/ms-vscode.node-debug/1.0.0/meta.yaml
benoitf: yes correct.
The broker is downloading plug-ins so I don't think it's an issue but it may be an improvement if not supported right now. BTW we can revisit later as well (like for GA)

l0rd:
@tolusha instead of having `ms-vscode.node-debug` and `ms-vscode.node-debug2` why don't you just:
```
ms-vscode.node-debug
      |_____1.32.1
      |      |_____meta.yaml
      |_____1.31.6
             |_____meta.yaml
```

tolusha:
@l0rd I know that it confuses a lot, but it isn't a node debugger with several different version. They are two different plugins that work together.

l0rd:
@tolusha ok thanks got it

tolusha:
The image [1] is available
[1] https://hub.docker.com/r/eclipse/che-theia-endpoint-runtime/tags

PR #81, opened 2019-01-31
Update versions of editors/plugin

garagatyi:
### What does this PR do?
Update versions of editors/plugin


skabashnyuk:
Should we upgrade a version of plugins and editors itself for such radical change? Wdyt @garagatyi @benoitf @l0rd ?

benoitf:
I think it will make extra work (update every stacks ?)

garagatyi:
@skabashnyuk No, we don't have plugins update policies for workspaces which mean that users would be stuck with old plugins.

garagatyi:
BTW UD shows all the versions of the same plugin. See
image
@slemeur FYI

PR #83, opened 2019-02-01
Update Theia editor and Exec plugin

garagatyi:
### What does this PR do?
Use direct links to che-plugin.yaml instead of tar.gz archives
to simplify release of new version.

### DIFF
--- a/plugins/org.eclipse.che.editor.theia/1.0.0/meta.yaml
+++ b/plugins/org.eclipse.che.editor.theia/1.0.0/meta.yaml
@@ -5,4 +5,4 @@ name: theia-ide
 title: Eclipse Theia for Eclipse Che
 description: Eclipse Theia
 icon: https://raw.githubusercontent.com/theia-ide/theia/master/logo/theia-logo-no-text-black.svg?sanitize=true
-url: https://github.com/ws-skeleton/che-editor-theia/releases/download/untagged-66e654fb4d0739982113/che-editor-plugin.tar.gz
+url: https://github.com/ws-skeleton/che-editor-theia/releases/download/untagged-cf91340b0b4af08ec756/che-editor-plugin.tar.gz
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.che.editor.theia/1.0.0/meta.yaml
```diff
@@ -5,4 +5,4 @@ name: theia-ide
 title: Eclipse Theia for Eclipse Che
 description: Eclipse Theia
 icon: https://raw.githubusercontent.com/theia-ide/theia/master/logo/theia-logo-no-text-black.svg?sanitize=true
-url: https://github.com/ws-skeleton/che-editor-theia/releases/download/untagged-66e654fb4d0739982113/che-editor-plugin.tar.gz
+url: https://raw.githubusercontent.com/ws-skeleton/che-editor-theia/f4b077eb01cd6373b0343f6029e5824348238a08/etc/che-plugin.yaml
```
benoitf: https://github.com/ws-skeleton/che-editor-theia/releases/download/untagged-cf91340b0b4af08ec756/che-editor-plugin.tar.gz ?

I don't like that hack around direct yaml link

garagatyi: In our case tar.gz just contains a single yaml. Does it make sense to package it as an archive?

benoitf: it's automatically published

garagatyi: Setting up automatic publishing in Eclipse repos is way trickier

benoitf: it's travis so AFAIK we have enabled travis on a lot of che repositories ?

garagatyi: Yes, but still requires the creation of issues in the bugzilla

garagatyi: But I'm OK to switch to archive if it is important for you

benoitf: would be great, thx

garagatyi: Updated the PR

PR #85, opened 2019-02-02
WIP - Adds VSCode-YAML plugin

JPinkney:
### What does this PR do?

This PR adds the [VSCode-YAML](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) plugin. 

The dockerfile for "jpinkney/che7-yaml" lives [here](https://github.com/JPinkney/Che7-YAML/blob/master/Dockerfile) though I'm not sure if we even need it and can just use the "wsskeleton/theia-endpoint-runtime" image.

Tested and working on che.openshift.io. However, only a subset of the features are working (documented [here](https://github.com/eclipse/che/issues/12142#issuecomment-452046280)).


tolusha:
Closed in favor of https://github.com/eclipse/che-plugin-registry/pull/97

PR #91, opened 2019-02-19
Parametarize `kubernetes.io/ingress.class`

monaka:
### What does this PR do?

Enable to customize the value of `kubernetes.io/ingress.class`.
This is related on eclipse/che#12645 and similar to eclipse/che#11820

### DIFF
--- a/kubernetes/che-plugin-registry/values.yaml
+++ b/kubernetes/che-plugin-registry/values.yaml
@@ -13,3 +13,5 @@ chePluginRegistryMemoryLimit: 256Mi
 
 global:
   ingressDomain: 192.168.99.100.nip.io
+  ingress:
+    class: 'nginx'
--- a/kubernetes/che-plugin-registry/templates/ingress.yaml
+++ b/kubernetes/che-plugin-registry/templates/ingress.yaml
@@ -12,7 +12,7 @@ kind: Ingress
 metadata:
   name: che-plugin-registry
   annotations:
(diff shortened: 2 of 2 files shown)


kubernetes/che-plugin-registry/values.yaml
```diff
@@ -13,3 +13,5 @@ chePluginRegistryMemoryLimit: 256Mi
 
 global:
   ingressDomain: 192.168.99.100.nip.io
+  ingress:
+    class: 'nginx'
```
sleshchenko: Please add new line at the end of file

monaka: I see. I'll do it soon.

kubernetes/che-plugin-registry/values.yaml
```diff
@@ -13,3 +13,5 @@ chePluginRegistryMemoryLimit: 256Mi
 
 global:
   ingressDomain: 192.168.99.100.nip.io
```
sleshchenko: @monaka Do you think it would be better to move it under `ingress` field. like
```yaml
global:
  ingress:
    domain: 192.168.99.100.nip.io
    class: 'nginx'
```
Or it would be better to leave as is for backward compatibility?

monaka: If I had to choice... it's better to rename `ingressDomain` in the another PR.
But I'm not picky about it.

PR #94, opened 2019-02-20
[Helm deployments] Enable TLS on the ingress.

monaka:
### What does this PR do?

Appends `tls:` object to the ingress if the secret name was specified.

### DIFF
--- a/kubernetes/che-plugin-registry/templates/ingress.yaml
+++ b/kubernetes/che-plugin-registry/templates/ingress.yaml
@@ -22,3 +22,9 @@ spec:
         backend:
           serviceName: che-plugin-registry
           servicePort: 8080
+{{- if .Values.chePluginRegistryIngressSecretName }}
+    tls:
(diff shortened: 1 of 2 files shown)


monaka:
@l0rd @sleshchenko Can I merge by myself this?

sleshchenko:
@monaka I think it is better to get at least one approve even for small changes

monaka:
@sleshchenko Thank you for your review.

And I agree with your opinion.
So I think it's better to be set more strict merge gate here same as `eclipse/che` repository. https://help.github.com/en/articles/enabling-required-reviews-for-pull-requests
(But I have no permission to set it..)

sleshchenko:
@monaka I think webmaster can do it of someone creates the corresponding issue https://bugs.eclipse.org. Could you register this task?

monaka:
@sleshchenko I see. Thanks for your suggest. I'll send the request there.

PR #96, opened 2019-02-20
Introduce vscode-java plugin

svor:
Signed-off-by: Valeriy Svydenko <vsvydenk@redhat.com>

Related issue: https://github.com/eclipse/che/issues/12436
### What does this PR do?
This PR introduces VS Code Java plugin [1]

[1] https://marketplace.visualstudio.com/items?itemName=redhat.java

### DIFF
--- a/plugins/org.eclipse.che.vscode-redhat.java/0.38.0/meta.yaml
+++ b/plugins/org.eclipse.che.vscode-redhat.java/0.38.0/meta.yaml
@@ -1,0 +1,14 @@
+id: org.eclipse.che.vscode-redhat.java
+version: 0.38.0
+type: VS Code extension
+name: Java Language Server
+title: Language Support for Java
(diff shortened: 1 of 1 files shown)


garagatyi:
@svor can you elaborate why you used URL instead of extension field?

svor:
@garagatyi Here is no special reason. What is the correct way?

benoitf:
I think the issue might be that by using the extension field there is no way to stick to a given version ?
like ?version=0.38.0

tolusha:
@benoitf 
It requires changes to vscode plugin broker.

svor:
@benoitf you are right! For now we know which version we want to use also `version` field in `meta.yaml` and parent folder has the same value

benoitf:
@tolusha yes I know, I was just trying to answer to Alex
Just saying it could be great to have that as enhancement

Issue #55, opened 2018-11-05
Naming convention


l0rd:
There is no proper naming convention for Che plugins yet. This is issue is to propose one. The goal is to make something simple but still powerful, more similar to the VS Code extensions naming convention than Java packages. 

- A plugin should be identified by 3 fields:
  - `publisher-id`: usually matches the github organisation or user
  - `plugin-id`: usually matches the github repository
  - `version`: in the form `MAJOR.MINOR.PATCH` and [using the Semantic Versioning](https://semver.org/)
- `publisher-id` and `plugin-id` should be lowercase and less than 64 chars (each). Only letters, digits and hyphen are allowed (LDH rule).
- A plugin should be go in the folder: `plugins/publisher_id/plugin_id/version/`

tsmaeder:
1. How does a person or organization get a publisher-id? 
2. Can you please give a concrete example of publisher-id?

benoitf:
We discussed as well to remove some stuff from yaml file (id, version) as it's duplicating the directory layout, should it be part of this enhancement or another one ?

benoitf:
@tsmaeder
1. as it's based on folders, if folder exists in https://github.com/eclipse/che-plugin-registry/tree/master/plugins then it is `taken` 
2. example of publisher id : `redhat`

tsmaeder:
>     1. as it's based on folders, if folder exists in https://github.com/eclipse/che-plugin-registry/tree/master/plugins then it is `taken`

So a committer decides whether to give a publisher-id to a certain when he accepts a PR with some folder name (like 'redhat')? What about squatters? How do we ensure that only redhatters can make 'redhat' plugins?

benoitf:
I would say it's up to reviewers. Also in meta.yaml you're referencing a repository. So probably if you try to propose redhat/cool-plugin but it ends to a repository http://github.com/thecoolguy/ then you can probably think that it's not part of a Red Hat project.

 I would say that this kind of features (Creating an account, publishing on its own namespace is for marketplace, here registry is just exposing the directory layout)

garagatyi:
After discussing with @l0rd and @ibuziuk we decided to move files in Che plugin registry according to the suggested FS structure. We also want to add model versioning to the structure, so it would be possible to change layout of meta files model in registry just by changing model version part in the path to the meta.yaml.
For example:
Now:
`plugin_ID/version/`
After this issue:
`/v1/plugins/publisher_id/plugin_id/version/`
[...]

PR #111, opened 2019-04-17
#55: Naming convention changes

garagatyi:
### What does this PR do?
Changes:
- Rename field name to displayName
- Rename field id to name
- Add v2 folder in root of registry. To use new notation registry URL needs to be changed to `old_url/v2/`
- Copy plugins into publishers folders under `/v2/` (but leave them in old location for backward compatibility)
- Remove /meta.yaml from plugin links in index to allow avoiding implicit addition of /meta.yaml to paths to meta files on downloading
- Add deprecate section to old plugins 
- Refactor scripts
- Add migrate section to registry index
- Add generated `id` field with `publisher/name/version` as a value to index
- Set publisher to plugins in new locations according to what we have discussed in #55 
- Set empty publisher in old plugins locations
- Remove dummy plugins with new notation under `/v2/` 

Related to https://github.com/eclipse/che-plugin-registry/issues/55


garagatyi:
@skabashnyuk @mkuznyetsov Please, review scripts - I changed that significantly.
@l0rd Please, review new plugins `publisher` and `name`

garagatyi:
@amisevsk @davidfestal @ibuziuk FYI

l0rd:
@garagatyi please move the new plugins in a `v2` folder (i.e. `v2/plugins/<publisher>/<name>/<version>`. Having old and new plugins in the same folder is kind of awkward. I don't want you to create a `v1` folder. We can leave the old plugins in the root folder but at least the `v2` folder will be clean and if we will continue using this pattern if when we introduce breaking changes (`v3`, `v4` etc...)

You should remove examples and dummy plugins as well (all in the `ws-skeleton` folder except the 2 editors).

README.md
```diff
@@ -58,10 +58,10 @@ Here is an overview of all fields that can be present in plugin meta YAML files
 
 |field name in YAML | description | mandatory |
 |-------|-------|-------|
-| id | plugin ID | yes |
+| name | plugin name | yes |
```
sleshchenko: it makes sense to describe format here. It should be URL compatible now, the same for publisher

garagatyi:
@l0rd I moved plugins with new notation to v2 folder.
To use old registry we should leave URL to the registry as is. This ensures that workspaces in older Che will work after the update. `id` in old index.json is equal to the name for backward compatibility.
To use new notation add `/v2` to the URL of the registry. This is what we will do in Che after updating UD and WS master/broker. When v2 is used `id` in index.json contains `publisher/name/version`.
I also removed dummy plugins as you suggested.

l0rd:
@garagatyi good job!

garagatyi:
Related https://github.com/eclipse/che/issues/12910

garagatyi:
@mkuznyetsov @skabashnyuk Please, review

garagatyi:
@l0rd If we remove dummy plugins existing workspaces with them, if any, would be broken. Is it OK for you?
Please, also review publisher/name in plugins to ensure there would not be any mistakes on my side.

garagatyi:
Take into account that old plugins we re also copied to v2 to avoid breaking existing workspaces that use old notation

garagatyi:
@l0rd I did not remove dummy plugins with old convention, only with new one. It did not break anything. Sorry for the misinformation

v2/plugins/redhat.vscode-yaml/0.3.0/meta.yaml
```diff
@@ -0,0 +1,17 @@
+publisher: ""
```
ibuziuk: redhat ?

v2/plugins/redhat.vscode-xml/0.3.0/meta.yaml
```diff
@@ -0,0 +1,17 @@
+publisher: ""
```
ibuziuk: redhat ?

v2/plugins/redhat.vscode-openshift-connector/0.0.17/meta.yaml
```diff
@@ -0,0 +1,17 @@
+publisher: ""
```
ibuziuk: redhat ?

README.md
```diff
@@ -94,30 +94,30 @@ Response:
 ```javascript
 [
   {
-    "id": "org.eclipse.che.editor.theia",
+    "name": "org.eclipse.che.editor.theia",
```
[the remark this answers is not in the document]
garagatyi: done

README.md
garagatyi: I added new fields. I didn't discuss URL compatibility constraints before, so leaving that to be contributed by others.

v2/plugins/redhat.vscode-openshift-connector/0.0.17/meta.yaml
garagatyi: It's old notation, publisher is empty for backward compatibility

v2/plugins/redhat.vscode-xml/0.3.0/meta.yaml
garagatyi: It's old notation, publisher is empty for backward compatibility

v2/plugins/redhat.vscode-yaml/0.3.0/meta.yaml
garagatyi: It's old notation, publisher is empty for backward compatibility

garagatyi:
@sleshchenko @ibuziuk I addressed your comments, please, take a look again

check_plugins_location.sh
```diff
@@ -8,39 +8,24 @@
...
-    VERSION_YAML=$(yq r meta.yaml version | sed 's/^"\(.*\)"$/\1/')
-    if [[ "$VERSION_YAML" != "$VERSION_DIR_NAME" ]];then
-      echo "!!! Version mismatch in plugin '${ID_DIR_NAME}/${VERSION_DIR_NAME}':"
-      echo "!!!   version in meta.yaml: '${VERSION_YAML}'"
-      echo "!!!   version directory name: '${VERSION_DIR_NAME}' "
+    expected_path_v1="./plugins/${plugin_id}/meta.yaml"
+    expected_path_v2="./v2/plugins/${plugin_id}/meta.yaml"
+    if [[ "${expected_path_v1}" != "$i"  && "${expected_path_v2}" != "$i" ]]; then
+      echo "!!! Location mismatch in plugin '${plugin_id}':"
+      echo "!!!   Expected location: '${expected_path}' or ${alternative_expected_path}"
```
mkuznyetsov: what is alternative_expected_path? I don't see it being assigned anywhere
also missing escaping quotes

garagatyi: You are looking at outdated diff

garagatyi: Sorry, my bad, I was on a wrong branch. I'll remove that. Thanks for catching

garagatyi:
It was difficult to properly rebase on top of master so I squashed all the commits. Sorry for the inconvenience.

README.md
```diff
@@ -58,84 +58,131 @@ Here is an overview of all fields that can be present in plugin meta YAML files
 
 |field name in YAML | description | mandatory |
 |-------|-------|-------|
-| id | plugin ID | yes |
+| publisher | publisher name | yes |
```
sleshchenko: I think it would be better to put `publisher`, `name`, `version` fields description together

README.md
```diff
@@ -58,84 +58,131 @@ Here is an overview of all fields that can be present in plugin meta YAML files
...
+| deprecate.migrateTo | which plugin should be used instead when plugin is deprecated | no |
```
sleshchenko: what is the format? `publisher`/`name`/`version`?

README.md
```diff
@@ -58,84 +58,131 @@ Here is an overview of all fields that can be present in plugin meta YAML files
...
+    }
+  },
+  {
+    "name": "theia",
+    "version": "next",
+    "publisher": "che-incubator",
+    "id": "che-incubator/theia/next",
     "type": "Che Editor",
-    "name": "theia-ide",
+    "displayName": "theia-ide",
```
sleshchenko: `Eclipse Theia` would be better as an example of `displayName` field

Dockerfile
```diff
@@ -8,11 +8,16 @@
 #
 FROM mikefarah/yq as builder
 RUN apk add --no-cache bash
-COPY .htaccess README.md /build/
+COPY .htaccess README.md check_plugins_location.sh check_plugins_images.sh check_plugins_viewer_mandatory_fields.sh index.sh index_v2.sh index_util.sh set_plugin_dates.sh util.sh /build/
```
sleshchenko: I wonder if we can copy all `sh` files without listing their names.

plugins/redhat.vscode-openshift-connector/0.0.17/meta.yaml
```diff
@@ -1,14 +1,17 @@
...
 icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
-publisher: Red Hat, Inc.
 repository: https://github.com/redhat-developer/vscode-openshift-tools
 category: Other
 firstPublicationDate: "2019-03-11"
 attributes:
   containerImage: "eclipse/che-remote-plugin-openshift-connector-0.0.17:next"
+deprecate:
+  autoMigrate: false
+  migrateTo: redhat/vscode-openshift-connector/0.0.17
```
sleshchenko: You have another plugin with the repository from `redhat-developer` organization and publisher there is `redhat-developer` https://github.com/eclipse/che-plugin-registry/pull/111/files#diff-eee18f0b9794223fd4f75a2007aeed2fR1
Here you use `publisher: redhat`.
Should publishers be the same for these plugins? If no - could you please elaborate more, when `redhat` and `redhat-developer` values should be used for publisher field?

v2/plugins/che-machine-exec-plugin/0.0.1/meta.yaml
```diff
@@ -0,0 +1,16 @@
...
+description: Che Plug-in with che-machine-exec service to provide creation terminal
+  or tasks for Eclipse CHE workspace machines.
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+url: https://github.com/eclipse/che-machine-exec/releases/download/0.0.2/che-service-plugin.tar.gz
+repository: https://github.com/eclipse/che-machine-exec/
+firstPublicationDate: "2019-02-05"
+category: Other
+deprecate:
+  autoMigrate: false
+  migrateTo: eclipse/che-machine-exec-plugin/0.0.1
```
sleshchenko: `migrateTo` is specified while `autoMigrate` is false. Could you please elaborate on how `migrateTo` is used in such case?

v2/plugins/che-machine-exec-plugin/0.0.1/meta.yaml
```diff
@@ -0,0 +1,16 @@
+publisher: ""
```
sleshchenko: Should not plugins in `v2` folder contain non-empty publisher?

sleshchenko: I got why it is empty here

README.md
garagatyi: yes

v2/plugins/che-machine-exec-plugin/0.0.1/meta.yaml
garagatyi: It is not used ATM, but I suppose that UD should show warning that plugin is deprecated and another one should be used instead. The same warning can be produced by the broker.

v2/plugins/che-machine-exec-plugin/0.0.1/meta.yaml
garagatyi: Unfortunately we need this for the backward compatibility. Otherwise, after an update of wsmaster existing workspaces with plugins in old notation would not start.

README.md
garagatyi: done

README.md
garagatyi: added

Dockerfile
garagatyi: done

plugins/redhat.vscode-openshift-connector/0.0.17/meta.yaml
garagatyi: I believe it changed URL of archive from VS Code marketplace to github. I didn't change publisher after that. Actually, I don't know what should we do in this case.

README.md
garagatyi: done

garagatyi:
FYI I reworked PR not to change anything in v1 plugins to ensure that everything will be backward compatible when OSIO registry will be updated

l0rd:
@garagatyi I still see a lot of plugins folders in the `v2` that are not in the format `<plublisher>/<name>` or that are sample plugins. I suspect that this is for backward compatibility but it doesn't make sense to me: we are in beta and I haven't seen any real adoption of Che 7 workspaces yet (except for IDE2 team I suppose). An email to che-dev explaining how to fix the config of an existing workspace should be enough.

garagatyi:
@l0rd yes, those are for the backward compatibility. 
I did think that backward compatibility is a concern. If it is not then we could simplify a lot of things.
@ibuziuk how do you think can we avoid adding backward compatibility for plugins notation for OSIO-Che?

l0rd:
@garagatyi @ibuziuk do you know any real user of Che 7 workspaces on osio-che? As a reminder Che 7 workspaces didn't worked there for a few months and stacks are really old ones.

ibuziuk:
@l0rd well, from the beginning, we were planning to have backward compatibility with existing workspaces on che.openshift.io. Are we sure that we opt for breaking things on the live service with just a che-dev email notification?

l0rd:
@ibuziuk my understanding is that Che 7 has been unusable on che-osio until 2 weeks ago. And that currently it's still kind of not easily usable. Hence I suspect that we don't have users and I would like to avoid that we spend time for non existing users. Backward compatibility makes sense only if we have users actively using Che 7 workspaces.

garagatyi:
@l0rd makes sense. My frustration comes from the fact that we didn't go to this direction initially. 
@ibuziuk if you are OK with the plan I'll go with it

ibuziuk:
well, I'm ok - it is just that I also thought that the backward compatibility was the initial plan we discussed during prio calls

garagatyi:
@l0rd but we will still leave `/plugins` and `/v2/plugins` in the registry because prod-preview deploys registry on each commit, so I don't want to break at least this

garagatyi:
I have just added discussed changes to the PR

garagatyi:
@l0rd please approve to ensure that now our vision are inlined

l0rd:
@garagatyi I agree with you about `v2/plugins` and `plugins` folders: in this case we want to keep the legacy `plugin` folder because we want to keep the compatibility with the current `ws-master`. But that's different from the existing che 7 workspaces configs backward compatibility. That may explain the misunderstanding.

PR #114, opened 2019-04-19
che #12840 Adding latest versions of yaml (0.4.0) and xml (0.5.1) vscode plugins

ibuziuk:
che #12840 Adding latest versions of yaml (0.4.0) and xml (0.5.1) vscode plugins
Also contains workaround for marketplace rate limmit error - https://github.com/eclipse/che/issues/12840


ibuziuk:
I'm not sure about `firstPublicationDate` which has been copied from the previous version of plugins

garagatyi:
@ibuziuk just put there current date

PR #116, opened 2019-04-19
Add shellcheck to Travis CI

garagatyi:
### What does this PR do?
Add shellcheck to Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,5 @@ sudo: required
 services: docker
 
 script:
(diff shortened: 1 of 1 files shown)


garagatyi:
My PR https://github.com/eclipse/che-plugin-registry/pull/111 brings a lot of changes, so I'm going to proceed with this PR only after merging that one

garagatyi:
Closing since Travis is so slow that we can barely use it. Instead of that I'll try that with CircleCI

PR #118, opened 2019-04-23
che #12840 Downloading sonarlint extention directly from the github release page

ibuziuk:
Workaround for https://github.com/eclipse/che/issues/12840
There is an inconsistency between version of extention in registry (0.0.1) and actual release (1.6.0), but we could probably treat it as a minor issue that should be tackled for the upcoming releases of the extention in the plugin registry future.


garagatyi:
@ibuziuk FYI CI is failing

ibuziuk:
@garagatyi seems fixed ;-)

benoitf:
FYI I restarted the CI job

PR #131, opened 2019-04-24
Using GH release instead of marketplace for go extension to avoid marketplace rate limit issues

ibuziuk:
### What does this PR do?
Related issue - https://github.com/eclipse/che/issues/12840


ibuziuk:
@garagatyi please review in the light of the naming convention changes

PR #133, opened 2019-04-25
 Fixing build for promototion to dsaas-stg. Adding build status of the main CI job https://ci.centos.org/job/devtools-che-plugin-registry-build-master

ibuziuk:
Adding build status of the main CI job https://ci.centos.org/job/devtools-che-plugin-registry-build-master 

Currently failing with:

> [91m/bin/sh: ./check_plugins_location_v1.sh: not found
> [0mThe command '/bin/sh -c cd plugins && ./check_plugins_location_v1.sh' returned a non-zero code: 127

### DIFF
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -12,8 +12,8 @@ COPY .htaccess README.md *.sh /build/
 COPY /plugins /build/plugins
 COPY /v2 /build/v2
 WORKDIR /build/
-RUN cd plugins && ./check_plugins_location_v1.sh
-RUN cd v2/plugins && ./check_plugins_location_v2.sh
+RUN ./check_plugins_location_v1.sh
+RUN ./check_plugins_location_v2.sh
 RUN ./check_plugins_images.sh
 RUN ./set_plugin_dates.sh
 RUN ./check_plugins_viewer_mandatory_fields_v1.sh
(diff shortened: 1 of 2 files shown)


Dockerfile.rhel
```diff
@@ -12,8 +12,8 @@ COPY .htaccess README.md *.sh /build/
 COPY /plugins /build/plugins
 COPY /v2 /build/v2
 WORKDIR /build/
-RUN cd plugins && ./check_plugins_location_v1.sh
-RUN cd v2/plugins && ./check_plugins_location_v2.sh
+RUN ./check_plugins_location_v1.sh
+RUN ./check_plugins_location_v2.sh
```
garagatyi: My bad, sorry. Can you also check whether it is needed to add `mkdir v2` where we do `mkdir plugins`?

garagatyi: 37th line

ibuziuk: no worries, well original Dockerfile is built just fine and does not contain `mkdir v2` but I have not tried to deploy in on minishift and verify that all is hosted fine. Should I ?

garagatyi: It contains `RUN mkdir /var/www/html/plugins` so I'm wondering whether we should add the same for v2

garagatyi: But this doesn't block this PR at all

ibuziuk: Ah, got it so you think both rhel and non-rhel might need to be updated with this fix `RUN mkdir /var/www/html/plugins/v2` since v2 might not be hosted without this change?

garagatyi: Non-rhel image works for me
Not sure about Rhel one

ibuziuk: ok. let's merge and see ;-)

PR #135, opened 2019-04-29
Merge meta.yaml and plugin yaml in v3

garagatyi:
### What does this PR do?
Merge meta.yaml and plugin yaml and put it to v3 folder as described [here](https://github.com/eclipse/che/issues/12908#issuecomment-474312103).
One thing that I changed comparing to the comment linked above is that now both theia and vscode plugins are using field `plugins` to refer to the archives of the plugins. 
Example of theia plugin:
```yaml
....
spec:
  plugins:
    - https://github.com/redhat-developer/omnisharp-theia-plugin/releases/download/v0.0.1/omnisharp_theia_plugin.theia
```
And example of vscode plugin:
```yaml
...
spec:
  plugins:
    - https://github.com/Azure/vscode-kubernetes-tools/releases/download/0.1.17/vscode-kubernetes-tools-0.1.17.vsix
```

@l0rd how do you think is it OK to refer to archives as `plugins` in both of them? Should we change to `extensions` for vscode and `plugins` for theia or should we use something completely different, like archives?

@tsmaeder FYI you can see in this PR how it will be possible to configure env vars, memory, volumes, etc for vscode extension in the registry. In this PR, in particular, VSCode extensions metas contain only container image, but since foramt is absolutely the same as with editors you can see how to do it in the theia meta, for example.


v3/plugins/redhat/java/0.38.0/meta.yaml
```diff
@@ -0,0 +1,18 @@
...
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/redhat-developer/vscode-java
+category: Language
+firstPublicationDate: "2019-02-20"
+spec:
+  containers:
+    - image: "eclipse/che-remote-plugin-runner-java8:next"
+  plugins:
+    - https://marketplace.visualstudio.com/_apis/public/gallery/publishers/vscjava/vsextensions/vscode-java-debug/0.16.0/vspackage
+    - http://download.jboss.org/jbosstools/static/jdt.ls/stable/java-0.38.0-1373.vsix
```
l0rd: No new line

garagatyi:
Makes sense. I'll go with field name `extensions`

v3/plugins/redhat/java/0.38.0/meta.yaml
garagatyi: Thanks for noticing, fixed

garagatyi:
@l0rd changed `plugins` to `extensions`

garagatyi:
@l0rd @ibuziuk please, review

v3/plugins/che-incubator/theia-dev/0.0.1/meta.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v2
```
ibuziuk: @garagatyi looks a bit misleading that `v3` plugins have `v2` API version. WDYT?

v3/plugins/ms-python/python/2019.3.6558/meta.yaml
```diff
@@ -0,0 +1,17 @@
...
+description: Linting, Debugging (multi-threaded, remote), Intellisense, code formatting, refactoring, unit tests, snippets, and more.
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/Microsoft/vscode-python
+category: Language
+firstPublicationDate: "2019-04-23"
+spec:
+  containers:
+    - image: "eclipse/che-remote-plugin-python-3.7.3:next"
+  extensions:
+    - https://github.com/Microsoft/vscode-python/releases/download/2019.3.6558/ms-python-release.vsix
```
ibuziuk: no new line in the end

garagatyi:
@ibuziuk we have different folders for different versions of the registry, but this doesn't mean those versions use different format of meta.yaml. 
The reason I put v2 as apiVersion is that it is second format we have that is not compatible with the current one, which i supposed to be v1. Naming convention didn't change format of the meta.yaml file, but did change placement of files.

v3/plugins/che-incubator/theia-dev/0.0.1/meta.yaml
garagatyi: answered below

v3/plugins/ms-python/python/2019.3.6558/meta.yaml
garagatyi: fixed

PR #143, opened 2019-05-09
Set up 'latest' dirs for each existing plugin

amisevsk:
### What does this PR do?
Adds script `setup_latest_dirs.sh` that is used in dockerfiles. This script looks through each plugin's directory for the 'latest' version of each plugin and copies that plugins meta.yaml to a new latest folder.

Rules for deciding which version is newest:
- If plugin already has a 'latest' folder, nothing is done (assumed to be managed manually)
- If plugin has a 'next' folder (e.g. che-theia) that meta.yaml is automatically used
- Otherwise, picks the meta.yaml with newest 'firstPublishedDate'

Only applies to v3 repo path.

Related: https://github.com/redhat-developer/rh-che/issues/1395


l0rd:
@amisevsk I am not sure it's a good idea to have that automated. I mean I may want to add a `nightly` plugin version but don't want it to be `latest`. And `next`=`latest` as well is not a good rule, we are using it right now but we will eventually move to something more stable when 7.0.0 is out.

benoitf:
I also think that next != latest (for che-theia next should be latest stable and not the current master branch for example)

amisevsk:
@benoitf Are we sure on that? We're currently not releasing versions of theia here, so `next` is the suggested version in general.

amisevsk:
Bit a force-push mess fixing shellcheck issues, but the latest changes are:

1. Don't automatically generate latest dirs, but leave the script in place so that it can be run manually
2. Update check script to ignore latest dirs (as those would fail the check)
3. Add symlinks to latest `meta.yaml`s in each `latest` folder.

@l0rd PTAL.

benoitf:
@amisevsk the 1.0.0 was used for latest (and it's strange as well as we update continuously 1.0.0 to the che-theia:latest image)

so AFAIK it's more the 1.0.0 version that should be considered as `latest`  as it is using the latest docker image and `next` which is using the `next` image

amisevsk:
@benoitf Makes sense; I'll update the PR.

amisevsk:
New changes:
- Added script to ensure every plugin directory contains a `latest` folder with `meta.yaml` inside it (necessary to enable support the wsmaster side).
- Moved all the `.sh` files into a `scripts` directory to clean up root folder a bit.

@ibuziuk @l0rd If you have a moment, please reapprove since this changes CI somewhat and could interfere with some of the open PRs we have.

amisevsk:
@benoitf Update since I almost forgot: to double check, I should update `che-theia/latest` to be `1.0.0` and not `next`? It looks like the `latest` image on dockerhub is a month old at this point, and we also have PR https://github.com/redhat-developer/rh-che/pull/1414 in rh-che to use `next` by default.

l0rd:
@amisevsk `v3/.../che-theia/1.0.0` should be be **replaced** (not linked) by `v3/.../che-theia/latest`. And that's because version `1.0.0` doesn't make sense: the meta.yml points to `latest`.

That said I would continue to use `che-theia/next` upstream and on rhche when no explicit editor is specified in a devfile for rc1. We should then switch to `latest` for rc2 and GA.

@benoitf are you ok with this approach?

amisevsk:
To add a blocker to this being merged: this results in plugins with version `latest` showing up twice in the user dashboard (since they're indexed twice). This is a TODO, along with the above comment.

amisevsk:
I'm working on the duplicate plugin issue mentioned above and it brings up a question about how we treat `theia/latest`.

- If we choose to replace (not link) the `meta.yaml`, it means that we treat `theia` differently from the other plugins in the registry
  - We could copy meta.yamls instead of linking for all plugins, although that means information is duplicated with no link
  - We could leave `theia` as a special case, and choose to support `latest` with no reference release (similar to docker)
- Letting `latest` *not* be a simple alias to a released version means that the indexing script needs to be significantly reworked -- instead of e.g. skipping latest directories we would have to dedupe the list of plugins in the index.

Another, alternate proposal: create copies of the 'latest' meta.yamls in the `latest` directory, and rewrite the version in those `meta.yamls` to be `latest`. This means managing these latest meta.yamls directly and copying in changes from other `meta.yaml`s manually, which is messy (can no longer easily compare `latest` meta with newest version), but it would solve the indexing and theia issues directly.

This is a significant downside of our approach to the registry as it is.

@l0rd any thoughts on the matter?

amisevsk:
Okay, version 3 of this PR.

New approach:
- For each plugin, `meta.yaml` is a separate file with version `latest`.
- The `setup_latest_dirs.sh` script has been rewritten to copy the newest meta.yaml into the latest directory
  - If no meta.yaml exists, it takes the newest one, sets its version to `latest`, and copies it
  - If a meta.yaml exists, it compares it to the 'newest' one on a json level
    - If they are different, it prints a warning to the user
    - If they are the same nothing is done
- `eclipse/che-theia/1.0.0` has been renamed to `che-theia/latest` (this may break compatibility with old workspaces -- is that acceptable?)
- We validate `latest` dirs in the same way as others -- they have to contain a plugin with version latest.

l0rd:
> eclipse/che-theia/1.0.0 has been renamed to che-theia/latest (this may break compatibility with old workspaces -- is that acceptable?)

I think that almost nobody will be impacted: those that use `v3` should use theia `next` but send an email to che-dev just in case before deploying it on `che-plugin-registry.openshift.io`

PR #144, opened 2019-05-13
Added definitions for gwt ide versions since 6.19.0

skabashnyuk:
### What does this PR do?
versions is used
- 6.19.0
- 7.0.0-beta-1.0
- 7.0.0-beta-2.0
- 7.0.0-beta-3.0
- 7.0.0-beta-4.0
- nightly


v3/plugins/eclipse/gwt/next/meta.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 publisher: eclipse
 name: gwt
-version: 1.0.0
+version: next
```
sleshchenko: Why next? Maybe it should be renamed accordingly to docker image version `nightly`?

v3/plugins/eclipse/gwt/6.19.0/meta.yaml
```diff
@@ -0,0 +1,85 @@
+apiVersion: v2
```
sleshchenko: What it apiVersion here? Should be it the same as root folder name? For this file it is `v3`

v3/plugins/eclipse/gwt/nightly/meta.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 publisher: eclipse
 name: gwt
```
sleshchenko: Consider renaming it to `gwt-ide` because `eclipse/gwt` says nothing that it is GWT based IDE for Eclipse Che.

v3/plugins/eclipse/gwt/6.19.0/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+publisher: eclipse
+name: gwt
+version: 6.19.0
+type: Che Editor
+displayName: gwt-ide
```
sleshchenko: Display name should be something like `Eclipse GWT IDE`

v3/plugins/eclipse/gwt/6.19.0/meta.yaml
sleshchenko: Now I see that it should be `v2` here. More see https://github.com/eclipse/che-plugin-registry/pull/135#issuecomment-488632994

PR #146, opened 2019-05-14
Set 3.4.0 tag version of Dirigible

skabashnyuk:
### What does this PR do?
Set 3.4.0 tag version of Dirigible instead of latest docker image. This can help to improve reproducibility/stability.


skabashnyuk:
CC @delchev

PR #150, opened 2019-05-16
Update plugin meta.yaml spec documentation

amisevsk:
### What does this PR do?
Updates the documentation around what we expect for `meta.yaml`s in the registry, to clarify issues around latest changes.

@l0rd Please double check I got the current spec correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,15 +61,15 @@ Here is an overview of all fields that can be present in plugin meta YAML files.
 
 ```yaml
 apiVersion:            # v2; v1 supported for backwards compatability
-publisher:             # publisher name
-name:                  # plugin name
-version:               # plugin version
+publisher:             # publisher name; must match [-a-z0-9]+
+name:                  # plugin name; must match [-a-z0-9]+
+version:               # plugin version; must match [-.a-z0-9]+
 type:                  # plugin type; e.g. "Theia plugin", "Che Editor"
 displayName:           # name shown in user dashboard
 title:                 # plugin title
 description:           # short description of plugin's purpose
 icon:                  # link to SVG icon
-repository:            # repo hosting plugin sources
+repository:            # URL for plugin (e.g. Github repo)
 category:              # see [1]
 firstPublicationDate:  # optional; see [2]
 latestUpdateDate:      # optional; see [3]
@@ -77,15 +77,15 @@ deprecate:             # optional; section for deprecating plugins in favor of o
   autoMigrate:         # boolean
   migrateTo:           # new plugin id
 spec:                  # spec (used to be che-plugin.yaml)
-  endpoints:           # optional; plugin endpoints
+  endpoints:           # optional; plugin endpoints -- see https://www.eclipse.org/che/docs/che-6/servers.html for more details
     - name:
-      public:            # boolean
+      public:            # if true, endpoint is exposed publicly
       targetPort:
       attributes:
-        protocol:        # protocol used for communicating over endpoint
-        secure:          # use secure version of protocol above
-        discoverable:    # if false, no service is created for this endpoint
-        cookiesAuthEnabled:
+        protocol:        # protocol used for communicating over endpoint, e.g. 'ws' or 'http'
+        secure:          # use secure version of protocol above; convert 'ws' -> 'wss', 'http' -> 'https'
+        discoverable:    # if false, no k8s service is created for this endpoint
+        cookiesAuthEnabled: # if true, endpoint is exposed through JWTProxy
         type:
         path:
   containers:          # optional; sidecar containers for plugin
@@ -96,8 +96,8 @@ spec:                  # spec (used to be che-plugin.yaml)
         - name:
           value:
       volumes:           # volumes required by plugin
-          - mountPath:
-            name:
+        - mountPath:
+          name:
       ports:             # ports exposed by plugin
         - exposedPort:
       commands:          # commands available to plugin container
--- a/README.md
+++ b/README.md
@@ -57,37 +57,73 @@ helm delete --purge che-plugin-registry
 docker run -it  --rm  -p 8080:8080 eclipse/che-plugin-registry
 ```
 ### Plugin meta YAML structure:
-Here is an overview of all fields that can be present in plugin meta YAML files
+Here is an overview of all fields that can be present in plugin meta YAML files. This document represents the current `v3` version.
 
-|field name in YAML | description | mandatory |
-|-------|-------|-------|
-| publisher | publisher name | yes |
-| name | plugin name | yes |
-| version | plugin version | yes |
-| deprecate.migrateTo | which plugin should be used instead when plugin is deprecated. Format: publisher/name/version | no |
-| deprecate.autoMigrate | if deprecated plugin should be automatically replaced by another plugin specified in deprecate.migrateTo | no |
-| type | plugin type | yes|
-| displayName | plugin name | yes|
-| title| plugin title | yes |
-| description| plugin description | yes |
-| icon| URL to plugin icon (must be in SVG format) | yes |
-| url| an URL to the plugin source | yes |
-| publisher| name of plugin publisher | yes |
-| repository | URL to repository of the plugin | yes |
-| category| plugin category | yes<sup>1</sup> |
-| firstPublicationDate | date of publishing the plugin (in ISO 8601) | no<sup>2</sup> |
-| latestUpdateDate | date of latest plugin update (in ISO 8601) | no<sup>3</sup> |
-| preview | a URL to devfile, to preview this plugin | no |
-| tags | a list of tags, related to this plugin | no|
-| mediaImage | links for images showcasing the plugin | no |
-| mediaVideo | links for video showcasing the plugin| no |
-| attributes | a map of special attributes, can be used for instance, imported plugins from VS Code | no |
+```yaml
+apiVersion:            # v2; v1 supported for backwards compatability
+publisher:             # publisher name
+name:                  # plugin name
+version:               # plugin version
+type:                  # plugin type; e.g. "Theia plugin", "Che Editor"
+displayName:           # name shown in user dashboard
+title:                 # plugin title
+description:           # short description of plugin's purpose
+icon:                  # link to SVG icon
+repository:            # repo hosting plugin sources
+category:              # see [1]
+firstPublicationDate:  # optional; see [2]
+latestUpdateDate:      # optional; see [3]
+deprecate:             # optional; section for deprecating plugins in favor of others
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
-| latestUpdateDate | date of latest plugin update (in ISO 8601) | no<sup>3</sup> |
-| preview | a URL to devfile, to preview this plugin | no |
-| tags | a list of tags, related to this plugin | no|
-| mediaImage | links for images showcasing the plugin | no |
-| mediaVideo | links for video showcasing the plugin| no |
-| attributes | a map of special attributes, can be used for instance, imported plugins from VS Code | no |
+Here is an overview of all fields that can be present in plugin meta YAML files. This document represents the current `v3` version.
+
+```yaml
+apiVersion:            # v2; v1 supported for backwards compatability
```
tsmaeder: What about V3?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+publisher:             # publisher name
+name:                  # plugin name
+version:               # plugin version
```
tsmaeder: The combination of publisher, name, version make up the id. Must they be globally unique?

tsmaeder: What's the format of the version string?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
-| preview | a URL to devfile, to preview this plugin | no |
-| tags | a list of tags, related to this plugin | no|
-| mediaImage | links for images showcasing the plugin | no |
-| mediaVideo | links for video showcasing the plugin| no |
-| attributes | a map of special attributes, can be used for instance, imported plugins from VS Code | no |
+Here is an overview of all fields that can be present in plugin meta YAML files. This document represents the current `v3` version.
+
+```yaml
+apiVersion:            # v2; v1 supported for backwards compatability
+publisher:             # publisher name
```
tsmaeder: Are there restrictions on the what can be contained here? Can the publisher contain stuff like '/'? Is there escaping when constructing the plugin id?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+name:                  # plugin name
+version:               # plugin version
+type:                  # plugin type; e.g. "Theia plugin", "Che Editor"
+displayName:           # name shown in user dashboard
+title:                 # plugin title
+description:           # short description of plugin's purpose
```
tsmaeder: How long, what should it contain?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+icon:                  # link to SVG icon
```
tsmaeder: Does this need to be a publicly available url? Does it still have to point to a *.svg?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+repository:            # repo hosting plugin sources
```
tsmaeder: What needs to be here? An url? A name?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+category:              # see [1]
+firstPublicationDate:  # optional; see [2]
+latestUpdateDate:      # optional; see [3]
+deprecate:             # optional; section for deprecating plugins in favor of others
+  autoMigrate:         # boolean
+  migrateTo:           # new plugin id
```
tsmaeder: Plugin id as in publisher/name/version? What if it's in a different plugin registry?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+spec:                  # spec (used to be che-plugin.yaml)
+  endpoints:           # optional; plugin endpoints
+    - name:
+      public:            # boolean
```
tsmaeder: What does it mean for an endpoint to be "public"?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+      targetPort:
+      attributes:
+        protocol:        # protocol used for communicating over endpoint
```
tsmaeder: possible values? format?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+        secure:          # use secure version of protocol above
+        discoverable:    # if false, no service is created for this endpoint
```
tsmaeder: What's a service?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+        cookiesAuthEnabled:
```
tsmaeder: What does this do?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+        type:
```
tsmaeder: Possible values?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+        path:
```
tsmaeder: Path of what?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+  containers:          # optional; sidecar containers for plugin
+    - image:
+      name:              # name used for sidecar container
+      memorylimit:       # Kubernetes/OpenShift-spec memory limit string (e.g. "512Mi")
+      env:               # list of env vars to set in sidecar
+        - name:
+          value:
+      volumes:           # volumes required by plugin
```
tsmaeder: Will this create volumes? How does it play with the volumes defined in a workspace config?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+          - mountPath:
+            name:
+      ports:             # ports exposed by plugin
```
tsmaeder: What's the relationship between "ports" and "endpoints"?

README.md
```diff
@@ -57,38 +57,74 @@ helm delete --purge che-plugin-registry
...
+          workingDir:
+          command:       # list of commands + arguments, e.g.:
+            - rm
+            - -rf
+            - /cache/.m2/repository
+      mountSources:      # boolean
+  workspaceEnv:        # optional; env vars for the workspace
+    - name:
+      value:
+  extensions:            # optional; required for VS Code/Theia plugins; list of urls to plugin artifacts (.vsix/.theia files) -- examples follow
```
tsmaeder: This is input to the plugin broker, right? How can the user find out which plugin file formats are supported? Shouldn't we mention the restriction on downloading from the vs code marketplace here?

README.md
amisevsk: There is no `v3`. If you are thinking about the `v3` path on the registry side, that is a separate version (and relates to registry design).

README.md
amisevsk: Semantically, a publisher cannot have two plugins with the same name and version, I don't believe this requires explicit documentation.

amisevsk: Anything you want, currently. We're currently moving to support `latest` as an alias to the newest version released.

README.md
amisevsk: I'll add explicit documentation here. I believe the requirement is alphanumeric + `.`, `-`, and `_`.

README.md
amisevsk: It should contain a description of what the plugin does; I don't think we have a limit on length.

README.md
amisevsk: Yes, new plugin ID. It doesn't make sense to support migrating to another registry IMO since it would be something we don't control. It's intended purpose is to allow a smoother transition between plugin ids *within* the registry. 

@l0rd can confirm.

README.md
amisevsk: Visible outside of the internal pod network, currently; will add doc.

README.md
amisevsk: I believe it does; would need to confirm elsewhere.

cc: @l0rd

README.md
amisevsk: Updated.

README.md
amisevsk: The openshift/kubernetes object.

amisevsk: Added k8s quantifier.

README.md
amisevsk: Added doc.

README.md
amisevsk: The meaning of this one is lost to time immemorial. Perhaps @sleshchenko can answer?

README.md
amisevsk: Should be API path but I can't track how it's used across the infrastructure backend.

README.md
amisevsk: No idea.

README.md
amisevsk: Ports are exposed on a container; endpoints are exposed separately. The analogue is container ports and service ports -- should be familiar to people using k8s/openshift.

README.md
amisevsk: There aren't currently many restrictions here; the broker will download whatever you give it. 

I hesitate to codify the VS Code marketplace limitation here but can add it if it is important.

README.md
amisevsk: Updated; URL to "plugin", generally is url of github repo, but could ostensibly be a link to your project homepage.

README.md
tsmaeder: Well, the fact that publisher/name/version is makes up the unique identifier is not obvious to a noob.

README.md
tsmaeder: How about describing this as "Version of the meta.yml file format" to prevent others from thinking the same as me.

README.md
tsmaeder: Thanks for the explanation, put it in the doc :-)

README.md
tsmaeder: "analogue" as in "will map to"? That would be useful info.

README.md
amisevsk: Similarly though, I would not expect a user be surprised when releasing two identical plugins does not work. Are there any marketplaces/registries that allow this?

README.md
amisevsk: Great point -- I've addressed a few of these comments elsewhere already. 

Will do :)

README.md
amisevsk: Added regexes for these; to close the loop: publisher and name are alphanumeric plus `'-'`, version can additionally contain `'.'`

amisevsk:
@tsmaeder Your review brings up a lot of great questions about the current `meta.yaml` spec (ones I'm not capable of answering entirely right now). I've created issue https://github.com/eclipse/che-plugin-registry/issues/152 to track additional documentation that needs to occur.

PR #158, opened 2019-05-28
Increase mem. limit for a Kubernetes Che Plugin sidecar to 1G

azatsarynnyy:
### What does this PR do?
Increases memory limit for a Kubernetes Che Plugin sidecar to 1G.
The amount of memory has been determined experimentally. Less memory causes unstable work of Helm Repos tree.

### DIFF
--- a/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/1.0.0/meta.yaml
+++ b/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/1.0.0/meta.yaml
@@ -13,5 +13,6 @@ firstPublicationDate: "2019-05-15"
 spec:
   containers:
     - image: "eclipse/che-remote-plugin-kubernetes-tooling-1.0.0:next"
+      memoryLimit: "1G"
   extensions:
     - https://github.com/Azure/vscode-kubernetes-tools/releases/download/1.0.0/vscode-kubernetes-tools-1.0.0.vsix


v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/1.0.0/meta.yaml
```diff
@@ -13,5 +13,6 @@ firstPublicationDate: "2019-05-15"
 spec:
   containers:
     - image: "eclipse/che-remote-plugin-kubernetes-tooling-1.0.0:next"
```
sleshchenko: As far as I understand this image can be used for multiple vscode extensions, do you know what will be memory limit for the container if multiple vs code extensions define different memory limit?

azatsarynnyy: > As far as I understand this image can be used for multiple vscode extensions

I don't know any except the K8s Plugin by now.

> do you know what will be memory limit for the container if multiple vs code extensions define different memory limit?

 I don't. Do you?
Good question BTW )

sleshchenko: > I don't know any except the K8s Plugin by now.

Ah, I see that it's a special docker image for k8s tooling. 

> I don't. Do you?

I don't.

PR #162, opened 2019-06-06
Use eclipse/che-remote-plugin-runner-java11 image for java based plugins

svor:
Signed-off-by: Valeriy Svydenko <vsvydenk@redhat.com>

### What does this PR do?
Changes container image in java based plugins to use Java11

Updated plugins:
- redhat-java
- redhat-xml
- apache-camel

### Related issue
https://github.com/eclipse/che/issues/13225

### DIFF
--- a/v3/plugins/sonarsource/sonarlint-vscode/0.0.1/meta.yaml
+++ b/v3/plugins/sonarsource/sonarlint-vscode/0.0.1/meta.yaml
@@ -12,6 +12,6 @@ category: Linter
 repository: https://www.sonarlint.org/
 spec:
   containers:
-    - image: "garagatyi/remotetheia:java"
+    - image: "eclipse/che-remote-plugin-runner-java11:next"
   extensions:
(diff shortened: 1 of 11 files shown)


v3/plugins/sonarsource/sonarlint-vscode/0.0.1/meta.yaml
```diff
@@ -12,6 +12,6 @@ category: Linter
 repository: https://www.sonarlint.org/
 spec:
   containers:
-    - image: "garagatyi/remotetheia:java"
+    - image: "eclipse/che-remote-plugin-runner-java11:next"
```
benoitf: I thought sonarlint was not working with new versions of java. It has been fixed?

svor: looks like you are right, in the documentation I see that requirement is Java Runtime (JRE) 8.
So, I've changed it back to use java 8. Thanks!

PR #168, opened 2019-06-17
Ensure all plugin containers have names

amisevsk:
### What does this PR do?
Adds names to all plugin containers. Without a name sidecar containers get a three-letter randomly generated string as a name which is very hard to parse when looking at a workspace.

cc: @JPinkney

### DIFF
--- a/v3/plugins/camel-tooling/vscode-apache-camel/0.0.14/meta.yaml
+++ b/v3/plugins/camel-tooling/vscode-apache-camel/0.0.14/meta.yaml
@@ -13,6 +13,7 @@ firstPublicationDate: "2018-04-23"
 spec:
   containers:
     - image: "eclipse/che-remote-plugin-runner-java11:next"
+      name: vscode-apache-camel
       memoryLimit: "512Mi"
   extensions:
     - https://github.com/camel-tooling/camel-lsp-client-vscode/releases/download/0.0.14/camel-tooling.vscode-apache-camel-0.0.14.vsix
--- a/v3/plugins/camel-tooling/vscode-apache-camel/latest/meta.yaml
+++ b/v3/plugins/camel-tooling/vscode-apache-camel/latest/meta.yaml
@@ -13,6 +13,7 @@ firstPublicationDate: '2018-04-23'
 spec:
   containers:
   - image: eclipse/che-remote-plugin-runner-java11:next
+    name: vscode-apache-camel
     memoryLimit: "512Mi"
   extensions:
   - https://github.com/camel-tooling/camel-lsp-client-vscode/releases/download/0.0.14/camel-tooling.vscode-apache-camel-0.0.14.vsix
--- a/v3/plugins/che-incubator/typescript/1.30.2/meta.yaml
+++ b/v3/plugins/che-incubator/typescript/1.30.2/meta.yaml
@@ -13,5 +13,6 @@ firstPublicationDate: "2019-02-19"
 spec:
   containers:
    - image: "eclipse/che-theia-endpoint-runtime:next"
+     name: "vscode-typescript"
   extensions:
    - https://github.com/che-incubator/ms-code.typescript/releases/download/v1.30.2/che-typescript-language.vsix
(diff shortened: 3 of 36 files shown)


v3/plugins/camel-tooling/vscode-apache-camel/0.0.14/meta.yaml
```diff
@@ -13,6 +13,7 @@ firstPublicationDate: "2018-04-23"
 spec:
   containers:
     - image: "eclipse/che-remote-plugin-runner-java11:next"
+      name: vscode-apache-camel
```
benoitf: hello, what does it mean when several plugins will share the same container ?

let say vscode-apache-camel and vscode-java are launched, I will see a container named "vscode-apache-camel" ?
I would prefer to name the container based on the image name than then plug-in's name as image/container may be shared

JPinkney: While I'm not sure specifically what we should do about plugins that share the same container, the reason it's necessary is to give the user a lot more information about which plugin is running in which container and I don't think using the image name truly conveys that.

It is very important specifically for the languages team because we are currently writing a document describing how to troubleshoot many of the languages in Che. If the user needs to access a terminal to see if a language server is running for example, it is a lot easier to say: `open the terminal in the container that is prefixed with the name vscode-yaml` then it is to say run `ps aux | grep node` on every container and then look to see if the YAML language server is running inside the container. Also, in this example, if the YAML language server failed to start then we wouldn't be able to get any information about which container it is supposed to be running in. Technically, if we used the image name we could do something like 'eclipse/che-remote-plugin-runner-node' but I don't think it conveys enough information, especially if you had multiple plugins with the same image name, its too hard to diagnose which container holds which plugin in that case.

IMO end users do not particularly care what image name is being used for their plugin (as long as the image contains everything it needs), rather they are interested in knowing which container their plugin was deployed.

benoitf: @JPinkney yes but if I'm using vscode-java and that the container is named vscode-apache-camel I will never think that my vscode-java plug-in is there ?

JPinkney: In the case where multiple plugins are running in the same container I don't think end user would know to look inside of vscode-apache-camel for java, I agree. In that case I'm not sure what to do.

On the one hand, you can make the container name be vscode-apache-camel/vscode-java but that naturally lends itself to massive names (especially if you had 3 or 4 plugins running in the same container). But that at least conveys all the info. So theres pros and cons to that approach.

On the other hand, you can use image names at that would at least give the user the idea of where to start looking, but it doesn't convey all information.

It just feels wrong writing in language troubleshooting document please run `ps aux | grep node` in all containers to check if the YAML language server is running in the container

amisevsk: @benoitf AFAIK we don't currently merge plugins to run in the same container; when I run a workspace with e.g. camel and vscode-java, I get two sidecar `eclipse/che-remote-plugin-runner-java11:next` containers running with names like `5e2` and `o8f`. 

I would presume that merging plugins to run in the same container would have to have some other sort of naming scheme in any case, as other plugins in the registry already specify a container name and we don't have any spec requirement (to my knowledge) that suggests not including one. 

This PR would be useful because the end-user should not care what image is used to run the plugins they have enabled -- as it stands the options would be to 1) know that the vscode-camel plugin runs in the `eclipse/che-remote-plugin-runner-java11:next` image, or 2) in the case of non-merged sidecars, search through all the `eclipse/che-remote-plugin-runner-java11:next` images running to find which one is hosting the plugin.

svor: @amisevsk Now we have just one plugin to run 2 vscode extensions in one container - it is java-plugin, it contains vscode-java and java-debug extensions.

amisevsk: @svor Ah I see; that use case is already covered by the PR, since the plugin specifies one container and two extensions -- these will run in a sidecar named `vscode-java` with some random chars appended. I was assuming that the discussion was moreso in the lines of taking two separate plugins (e.g. vscode-camel and vscode-java) and running them in the same container since they use the same image. I'm not fully aware of the status of that at this moment.

PR #174, opened 2019-06-24
Fix OOM error for OpenShift Connector plugin

azatsarynnyy:

### What does this PR do?
https://github.com/eclipse/che/issues/13609

### DIFF
--- a/v3/plugins/redhat/vscode-openshift-connector/0.0.21/meta.yaml
+++ b/v3/plugins/redhat/vscode-openshift-connector/0.0.21/meta.yaml
@@ -14,5 +14,6 @@ spec:
   containers:
     - image: "eclipse/che-remote-plugin-openshift-connector-0.0.21:next"
       name: "vscode-openshift-connector"
+      memoryLimit: "512Mi"
   extensions:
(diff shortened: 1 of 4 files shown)


v3/plugins/redhat/vscode-openshift-connector/0.0.21/meta.yaml
```diff
@@ -14,5 +14,6 @@ spec:
   containers:
     - image: "eclipse/che-remote-plugin-openshift-connector-0.0.21:next"
       name: "vscode-openshift-connector"
+      memoryLimit: "512Mi"
```
sleshchenko: I'm not sure what is our strategy of supporting previous versions of plugins, but in this particular case looks like it cost nothing if we set memory limit for others `0.0.17` and `0.0.19` versions but it may prevent OOM if somebody uses not the latest plugin in his Devfile.

azatsarynnyy: makes sense
the PR has been updated

sleshchenko: :+1:

PR #177, opened 2019-07-03
Fix image version to the 7.0.0-rc-3.0

vparfonov:
Signed-off-by: Vitalii Parfonov <vparfono@redhat.com>

### What does this PR do?

Fix image version to the 7.0.0-rc-3.0

### DIFF
--- a/v3/plugins/camel-tooling/vscode-apache-camel/7.0.0-rc-3.0/meta.yaml
+++ b/v3/plugins/camel-tooling/vscode-apache-camel/7.0.0-rc-3.0/meta.yaml
@@ -1,0 +1,19 @@
+apiVersion: v2
+publisher: camel-tooling
+name: vscode-apache-camel
+version: latest
+type: VS Code extension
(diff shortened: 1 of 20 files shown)


benoitf:
hello, I'm not sure that it's the expected behaviour. 

If it expected then I don't see why we have different repositories / lifecycles / plug-ins

benoitf:
from what i've understood, :next of endpoints image needs to be tagged with the associated che-theia image but the plug-in version can stay the same

After it's tagged, then version are backed to :next 

ppl deploying the registry 7.0.0-RC3 will then have theia endpoint images of 7.0.0-RC3 

so no need to change stacks etc for each che version

vparfonov:
so i mean in meta.yaml always will be latest version like here? 
https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/ms-python/python/latest/meta.yaml#L4

benoitf:
The registry is already per che version, so no need to change plugin's versions. Just the image used by plug-ins

PR #178, opened 2019-07-04
release Che 7.0.0 rc 3.0

vparfonov:
### What does this PR do?

TODO

### DIFF
--- a/plugins/camel-tooling.vscode-apache-camel/0.0.14/meta.yaml
+++ b/plugins/camel-tooling.vscode-apache-camel/0.0.14/meta.yaml
@@ -11,4 +11,4 @@ repository: https://github.com/camel-tooling/camel-lsp-client-vscode
 category: Language
 firstPublicationDate: "2018-04-23"
 attributes:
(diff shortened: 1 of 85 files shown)


l0rd:
This PR makes sense but what it's not clear is how we are going to tag the image 7.0.0-rc-3.0.

`cico_build.sh` is used to build the image and push it to:

```
quay.io/openshiftio/che-plugin-registry:$GIT_COMMIT
and
quay.io/openshiftio/che-plugin-registry:latest
```

But we need to have a mechanism to tag it as `quay.io/openshiftio/che-plugin-registry:7.0.0-rc-3.0` as well. Do you have an idea already how we will do that?

rhopp:
I think the change to cico_build.sh have to me made as part of this PR. Or are we pushing this registry image also somewhere else?

vparfonov:
Looks like wee need to create new or improve this job https://ci.centos.org/job/devtools-che-plugin-registry-build-master on CentOS CI. Which have ability to push tag image

PR #180, opened 2019-07-09
Separate directory for VSCode extensions

vitaliy-guliy:

### What does this PR do?

For `eclipse/che-theia/7.0.0-rc-3.0` and `eclipse/che-theia/next`
- defines `VSCODE_PLUGINS` environment variable. It allow to install VS Code extensions to this directory.
- persists this directory
- adds `/vscode-plugins` to plugin sources
- persists `/home/theia/.theia` directory to be able to store Theia settings and setting of some plugins/extensions

Is needed for https://github.com/eclipse/che/issues/13834

### DIFF
--- a/v3/plugins/eclipse/che-theia/7.0.0-rc-3.0/meta.yaml
+++ b/v3/plugins/eclipse/che-theia/7.0.0-rc-3.0/meta.yaml
@@ -51,7 +51,9 @@ spec:
      image: eclipse/che-theia:7.0.0-rc-3.0
      env:
          - name: THEIA_PLUGINS
-           value: local-dir:///plugins
+           value: local-dir:///plugins,local-dir:///vscode-plugins
+         - name: VSCODE_PLUGINS
+           value: /vscode-plugins
          - name: HOSTED_PLUGIN_HOSTNAME
            value: 0.0.0.0
          - name: HOSTED_PLUGIN_PORT
@@ -59,6 +61,10 @@ spec:
      volumes:
          - mountPath: "/plugins"
            name: plugins
+         - mountPath: "/vscode-plugins"
+           name: vscode-plugins
+         - mountPath: "/home/theia/.theia"
+           name: theia-data
      mountSources: true
      ports:
          - exposedPort: 3100
--- a/v3/plugins/eclipse/che-theia/next/meta.yaml
+++ b/v3/plugins/eclipse/che-theia/next/meta.yaml
@@ -51,7 +51,9 @@ spec:
      image: eclipse/che-theia:next
      env:
          - name: THEIA_PLUGINS
-           value: local-dir:///plugins
+           value: local-dir:///plugins,local-dir:///vscode-plugins
+         - name: VSCODE_PLUGINS
+           value: /vscode-plugins
          - name: HOSTED_PLUGIN_HOSTNAME
            value: 0.0.0.0
          - name: HOSTED_PLUGIN_PORT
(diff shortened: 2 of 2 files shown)


sunix:
@vitaliy-guliy could you explain why this is needed ?
It means that we are persisting informations about installed plugins in volumes and not devfile.
what if i want to recreate a new workspace from the same devfile ? Won't have the plugin/extension i have installed previously ?

vitaliy-guliy:
@sunix This is needed to persist VS Code extensions that has been installed by a command given from VS Code marketplace. Yes, the installed extension is not stored to the devfile and you cannot recreate your workspace with this extension. It works only for the current workspace. For now we don't have a solution how to describe such kind of plugins in devfile directly without defining external meta.yaml

sunix:
> @sunix This is needed to persist VS Code extensions that has been installed by a command given from VS Code marketplace. Yes, the installed extension is not stored to the devfile and you cannot recreate your workspace with this extension. It works only for the current workspace. For now we don't have a solution how to describe such kind of plugins in devfile directly without defining external meta.yaml

plz make a proposal so we can do it in the devfile in a new issue

Something like: 
```yaml
id: 'link to the vscode ext binary',
type: vscode-ext
```

l0rd:
@vitaliy-guliy is this really related to https://github.com/eclipse/che/issues/13747. Because you talked about making the panel work with the new devfile model and it doesn't look like this PR address that.

vitaliy-guliy:
@l0rd yes, you are right. It's a minor change and it was decided to do it withing fixing the plugins panel.
I have created a [dedicated issue](https://github.com/eclipse/che/issues/13834) for that and fixed the PR description

l0rd:
@evidolob can you review this PR please?

vitaliy-guliy:
@amisevsk we don't need to change the plugin broker. This option is only for installing a vscode extension by a command running inside Theia.
Let's say, if you want to insall Ruby extension https://marketplace.visualstudio.com/items?itemName=rebornix.Ruby you have to: 
- open a marketplace
- copy an installation command `ext install rebornix.Ruby`, insert it to text/command field in the Plugin view
- in case there is no installation command, you can easily drag this button to the Plugins view

There is a code in Theia using this environment variable.
https://github.com/theia-ide/theia/blob/master/packages/plugin-ext-vscode/src/node/plugin-vscode-resolver.ts#L45

If the variable is set, Theia will install a vscode extension to predefined directory. In our case (in this PR), this directory is also added to volumes. This is some kind of persistence to not lost the extensions when restarting.

See my demo video explaining the feature https://youtu.be/2_GPZb1_Hew?t=329

nickboldt:
Is this PR still relevant? Tempted to close as out of date.

vitaliy-guliy:
It depends on what we prefer.

If we prefer to have a configuration in the devfile only, then we don't need this PR anymore. But then we have to create and issue to make it possible to specify/describe the vscode extension in the devfile directly without creating a separate meta.yaml.

If it's ok to have an ability to store the installed vscode extension in the persisted volume only, let's update this PR and merge.

As for me, I prefer the first option.

Anyway, we have already implemented the feature with extensions installing. But the extension will not be deployed immediately, only after restarting the workspace. But without the dedicated persisted directory you will not see your plugin.

I demonstrated it several times on the sprint review.
One from the videos https://www.youtube.com/watch?v=28WqjC-9ciQ

sunix:
If we are not about to merge that one let's close it and create/update epic on how to properly install vscode extensions on the editor or sidecars container, dynamically (if possible) or with workspace restart and with devfile update. Could you prepare that @vitaliy-guliy ?

ericwill:
Going to close this one for now, we can revisit this discussion at a later date.

PR #183, opened 2019-07-15
Set 7.0.0-next version of Che Theia image 

vparfonov:
Signed-off-by: Vitalii Parfonov <vparfono@redhat.com>

### What does this PR do?

According to the https://github.com/eclipse/che-theia/pull/347 need to use new Che Theia image


benoitf:
should it be instead 7.0.0-RC4-SNAPSHOT something like 7.0.0-nightly or 7.0.0-next ?
AFAIK images have never used the tag SNAPSHOT but nightly/latest/next

benoitf:
also it's unclear how release are made from che-theia branch, I didn't see update on https://github.com/eclipse/che/issues/13782

with SNAPSHOT name I don't know if it is on each commit, every night, etc

vparfonov:
> should it be instead 7.0.0-RC4-SNAPSHOT something like 7.0.0-nightly or 7.0.0-next ?
> AFAIK images have never used the tag SNAPSHOT but nightly/latest/next

see here https://github.com/eclipse/che-theia/pull/347/files#diff-0f90e09c9296ab5840fb280db0cad7c6R17 please

benoitf:
yes I see where it was but it didn't ring the bell when I saw it.

PR #187, opened 2019-07-22
Helm chart deployment

sparkoo:
### What does this PR do?

- fix helm chart deployment config
- fix image to `quay.io` one


sparkoo:
> Same remark as for the devfile-registry: we should use `apiVersion: apps/v1` for k8s Deployments

@l0rd done

sparkoo:
@l0rd same here. I don't have permissions, so feel free to merge if you're happy with changes.

PR #192, opened 2019-07-25
Add release script

l0rd:
Release script for plugin-registry


cico_release.sh
```diff
@@ -0,0 +1,102 @@
...
+export SCRIPT_DIR
+
+# Source environment variables of the jenkins slave
+# that might interest this worker.
+function load_jenkins_vars() {
+  if [ -e "jenkins-env.json" ]; then
+    eval "$(./env-toolkit load -f jenkins-env.json \
+            DEVSHIFT_TAG_LEN \
+            QUAY_ECLIPSE_CHE_USERNAME \
+            QUAY_ECLIPSE_CHE_PASSWORD \
```
ibuziuk: @l0rd you need to add different set of credentials for pushing to rhel to `https://quay.io/organization/openshiftio/`
cico job has been already patched - https://github.com/openshiftio/openshiftio-cico-jobs/pull/1048

ibuziuk: see https://github.com/eclipse/che-devfile-registry/blob/master/cico_build.sh#L23-L26

cico_release.sh
```diff
@@ -0,0 +1,102 @@
...
+  local TARGET=$1
+  docker tag "${IMAGE}" "$TARGET"
+  docker push "$TARGET"
+}
+
+function release() {
+  TARGET=${TARGET:-"centos"}
+  REGISTRY="quay.io"
+
+  if [ "$TARGET" == "rhel" ]; then
```
ibuziuk: this approach will not work for release since the script is running only one time and there is no `target` in place. I believe you need to build & push `centos` first, than login with another creds and push to `rhel`

l0rd:
@ibuziuk thanks for the feedbacks. I had not finished the PR yet sorry. I have opened a couple of related PRs ([cico side](https://github.com/openshiftio/openshiftio-cico-jobs/pull/1049) and [devfile-registry side](https://github.com/eclipse/che-devfile-registry/pull/63) and they are ready for review now. 

But there is no hurry. I am on PTO tomorrow and I prefer to merge it on monday.

The only thing I haven't addressed is something you mentioned and I think we should do: always tag using the commit hash as well. That may be done in separate PRs.

cico_release.sh
ibuziuk: actually, it looks like `extra_target: rhel` in the job definition could do the trick, so the comment above is probably not relevant

l0rd:
@rhopp @skabashnyuk @vparfonov any of you guys can please review/approve and unblock this PR? It has already been approved by @ibuziuk but he is not a code owner.

PR #195, opened 2019-07-26
WIP add fetch_resources.sh script to pull vsix and svg files [DO NOT MERGE]

nickboldt:
add fetch_resources.sh script to pull vsix and svg files into the plugin registry container (for airgap scenario)

Change-Id: I84a53235ab0fd9bd794e43c2273dc1b9f3aea922


nickboldt:
After building this image, I pushed it to quay.io/nickboldt/airgap-che-plugin-registry:nightly and then deployed it to my minishift.

However, when I tried to launch a workspace using the JDK 11 devfile I got this error:

```
Press F5 or click here to try again.
Cleaning /plugins dir
Starting Init Plugin Broker
Unified Che Plugin Broker
List of plugins and editors to install
- redhat/java/latest - Java Linting, Intellisense, formatting, refactoring, Maven/Gradle support and more...
- eclipse/che-machine-exec-plugin/0.0.1 - Che Plug-in with che-machine-exec service to provide creation terminal or tasks for Eclipse CHE workspace machines.
- eclipse/che-theia/7.0.0-next - Eclipse Theia, get the latest release each day.
Starting Che plugins and editor processing
Starting VS Code and Theia plugins processing
Downloading VS Code extension for plugin 'redhat/java/latest'
Get https://0.0.0.0:8080/resources//github.com/microsoft/vscode-java-debug/releases/download/0.19.0/vscode-java-debug-0.19.0.vsix: dial tcp 0.0.0.0:8080: connect: connection refused
Error: Failed to run the workspace: "Plugins installation process failed. Error: Plugin broking process for workspace workspacebd99wa1ukp7nrsh1 failed with error: Get https://0.0.0.0:8080/resources//github.com/microsoft/vscode-java-debug/releases/download/0.19.0/vscode-java-debug-0.19.0.vsix: dial tcp 0.0.0.0:8080: connect: connection refused"
```

So... I need to figure out how to reference the route for the running plugin registry container dynamically when the pluginbroker sees 0.0.0.0:8080 (or some other placeholder we can put in instead).

skabashnyuk:
As an intermediate solution, it could be, maybe (what problem we are trying to solve?). However, in general, I'm -1 on this pr.
Two arguments.
- Architecture: Plugin-registry is tent to be a metadata microservice.  Not a hosting of artifacts. 
  The idea with links to artifact was completely opposite to this pr.  Plugin authors can update them separately without the need to change anything in the plugin-registry side.
  
Legal. 
- I have some doubts that downloading and redistributing all of these artifacts are legal from the Eclipse Foundation point of view.

slemeur:
@skabashnyuk : The problem that nick is trying to solve is a deployment one. Imagine an airgap environment with no access to internet. How someone can get access to the plugins along with the installed Che and consume them without requiring access to internet. (Like a company relying on a private maven repository).
What would be your proposal?

skabashnyuk:
>What would be your proposal? 

customized plugin registry where the person who builds it takes all responsibility

l0rd:
@nickboldt the hostname is wrong. The plugin registry and the workspace are running in separate Pods: the plugin broker is requething a file at `0.0.0.0:8080` (local network interface) but no process is listening there (every Pod in kubernetes uses a different network interface). To fix that you could use the service name ([`che-plugin-registry`](https://github.com/eclipse/che-plugin-registry/blob/master/openshift/che-plugin-registry.yml#L68) and port 80) but that works only if the `plugin-registry` and the che workspaces are deployed in the same Kubernetes namespace (can be true for local Che installations but it's usually false in production deployments). So the only reliable solution here is to use a separate script, to run at runtime, and that updates the hostname using the [route](https://github.com/eclipse/che-plugin-registry/blob/master/openshift/che-plugin-registry.yml#L77) hostname (something you don't know at build time).

Anyway, regarding @slemeur and @skabashnyuk comments, this is an interesting option for users that want to build images for offline usage but it should not be the default option. The official images (`quay.io/eclipse/che-plugin-registry`) should not contain the plugins but only the metadata.

nickboldt:
> At this stage (it looks like a PoC, not a PR), and given the fact that we are in feature-freeze mode, this PR should not be considered for merging at all.

100% agree, this isn't for merging and it shouldn't be default behaviour. it's a PoC that will be needed in downstream CRW 2, but if the option to build this way can be turned on via a flag in upstream, so much the better. I'll cross that bridge once I solve the URL prefix problem first. :)

l0rd:
@nickboldt I think that's a use case that has sense and should be implemented upstream.

amisevsk:
Route issues should be resolved by https://github.com/eclipse/che/issues/14163, with some additional setup on the registry side.

nickboldt:
Out of date, see https://github.com/eclipse/che-plugin-registry/pull/205

PR #202, opened 2019-08-14
Add more memory for K8s plugin sidecar to avoid OOM

azatsarynnyy:

### What does this PR do?

Fixes https://github.com/eclipse/che/issues/14210 by increasing memory limit for K8s plugin sidecar.

### DIFF
--- a/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17/meta.yaml
+++ b/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17/meta.yaml
@@ -14,5 +14,6 @@ spec:
   containers:
     - image: "eclipse/che-remote-plugin-kubernetes-tooling-0.1.17:7.0.0-next"
       name: "vscode-kubernetes-tools"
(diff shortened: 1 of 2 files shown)


v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17/meta.yaml
```diff
@@ -14,5 +14,6 @@ spec:
   containers:
     - image: "eclipse/che-remote-plugin-kubernetes-tooling-0.1.17:7.0.0-next"
       name: "vscode-kubernetes-tools"
+      memoryLimit: "1G"
```
sleshchenko: What was the previous default value?
`1G` looks more than is needed for toolings that communicate with K8s/OS clusters. I may be wrong )

azatsarynnyy: I checked it with ~700 mb several months ago and it was OOM Killed

PR #205, opened 2019-08-15
Replace mikefarah/yq with kislyuk/yq

amisevsk:
### What does this PR do?
Replaces [`mikefarah/yq`](https://github.com/mikefarah/yq) with [`kislyuk/yq`](https://github.com/kislyuk/yq) and reworks scripts to accomodate this change:

- Instead of manually checking `meta.yaml`s for required fields, uses `jsonschema` to validate against a schema file
    - This also removes the requirement for the `check_plugins_images.sh` script
    - This process is quite slow, taking around 25 seconds for all our current metas
- The `index.sh` script is greatly simplified due to the capabilities of `yq`
- There are a few workarounds in place because invoking `yq` in a loop is tremendously slow
    - To get plugin ids for reporting purposes (i.e. in `util.sh`), I'm using `grep` instead of `yq` (replacing this function with `yq` adds ~30 seconds to the build time). This could fail if any of the fields are not on the same line
- There are also a couple workarounds since `yq` doesn't support in-place editing: 
    - To set `latestUpdateDate`, we use `sed` to append a line to each file.
    - I'm not clear on what the `latestUpdateDate` field is for, since it's [discarded by the plugin broker](https://github.com/eclipse/che-plugin-broker/blob/21952b6098bd8edab883c290561f6e1cd08d22da/brokers/unified/broker.go#L295)


Dockerfile
```diff
@@ -6,15 +6,15 @@
...
-RUN apk add --no-cache bash
-COPY .htaccess README.md ./scripts/*.sh /build/
+FROM alpine:3.10 AS builder
+RUN apk add --no-cache py-pip jq bash && pip install yq jsonschema
+
+COPY .htaccess README.md ./scripts/*.sh ./scripts/meta.yaml.schema /build/
 COPY /v3 /build/v3
 WORKDIR /build/
 RUN ./check_plugins_location.sh v3
-RUN ./check_plugins_images.sh
```
nickboldt: why did you remove check_plugins_images script?

Dockerfile.rhel
```diff
@@ -6,15 +6,15 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #
-FROM mikefarah/yq as builder
-RUN apk add --no-cache bash
-COPY .htaccess README.md ./scripts/*.sh /build/
+FROM alpine:3.10 AS builder
```
nickboldt: Do we use this file? Why not delete it?

scripts/check_plugins_images.sh
```diff
@@ -1,34 +0,0 @@
-#!/bin/bash
```
nickboldt: why delete this check script?

amisevsk: Because I included the check it does in the schema, so it's covered by the regular validation. Sorry that's not clear at all, should have included a comment.

Dockerfile.rhel
amisevsk: I'll create an issue for moving to ubi8, I don't think alpine is necessary, moreso convenient.

The only concern would be due to the difference in image size, since the broker has to be pulled for every workspace start, but I wonder how much of an issue this would actually be considering it would probably be cached pretty quickly.

amisevsk: Ah nevermind I see you've beat me to the punch: https://github.com/eclipse/che/issues/14113

I'd do it as a separate PR as I don't think it's a simple drop-in replacement.

scripts/check_plugins_images.sh
amisevsk: For reference, it should be [this line](https://github.com/eclipse/che-plugin-registry/pull/205/files#diff-672f4dbb06abfc6f01b3ac2e267a3e4cR29)

amisevsk:
Rebased on master, fixed the shellcheck issue (can't put anything after a directive it seems), and removed an errant paste in the schema.

amisevsk:
Relaxed the validation requirement for version numbers to include "number", since the version for `che-incubator/cpptools/0.1` parses as a number and not a string (not an issue for the other plugins, as e.g. `7.0.0` cannot be a number).

nickboldt:
+1 for this solution. Make it sew, number one.

PR #206, opened 2019-08-19
che #14283: Switching all the versions from 'next' to '7.0.0-next'

ibuziuk:
Signed-off-by: Ilya Buziuk <ibuziuk@redhat.com>

### What does this PR do?
**DO NOT MERGE** until `eclipse/che-machine-exec:7.0.0-next` is released 
 Switching all the versions from 'next' back to '7.0.0-next' aka `find . -name '*.yaml' -exec sed -i -e 's/next/7.0.0/g' {} \;`
Basically, this PR reverts https://github.com/eclipse/che-plugin-registry/pull/203 
currently `next` version of the plugins is not compatible with Che master.
related issues:
 - https://github.com/eclipse/che/issues/14283
 -  https://github.com/redhat-developer/rh-che/issues/1565

### DIFF
--- a/v3/plugins/camel-tooling/vscode-apache-camel/0.0.14/meta.yaml
+++ b/v3/plugins/camel-tooling/vscode-apache-camel/0.0.14/meta.yaml
@@ -12,7 +12,7 @@ category: Language
 firstPublicationDate: "2018-04-23"
 spec:
   containers:
-    - image: "eclipse/che-remote-plugin-runner-java11:next"
+    - image: "eclipse/che-remote-plugin-runner-java11:7.0.0-next"
       name: vscode-apache-camel
       memoryLimit: "512Mi"
   extensions:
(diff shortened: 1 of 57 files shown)


ibuziuk:
@vparfonov for the sake of consistency can we release `eclipse/che-machine-exec:7.0.0-next` ? (I believe it should be the same as `7.0.0` for now)

ibuziuk:
@vparfonov I have also added back `7.0.0-next` version of che-theia to avoid backward compatibility isssues. I believe we should never remove existing versions of plugins in master

vparfonov:
> @vparfonov for the sake of consistency can we release `eclipse/che-machine-exec:7.0.0-next` ? (I believe it should be the same as `7.0.0` for now)

what the reason use `7.0.0-next`? Will be easy just set to the 7.0.0. Also in master we need to use next version as a rule

ibuziuk:
> what the reason use 7.0.0-next?

For consistency since all the plugins in this PR use `7.0.0-next` as a version of the runtime containers

> Also in master we need to use next version as a rule

yes, but only if `next` version works correctly, which is not the case. Otherwise, we would face issues similar to:
- eclipse/che#14283
- redhat-developer/rh-che#1565

PR #208, opened 2019-08-21
[airgap] explicitly reference docker.io registry

nickboldt:
explicitly reference docker.io registry since that's where all the images live; also let's be consistent with double quotes, single quotes, and no quotes: switch to quotes everywhere

Change-Id: If528213681ab79ac60b4fd97839b8f275bc56de9


nickboldt:
Purpose of this change is to make it easier to replace 3rd party repo references w/ airgapped customer private repo reference via sed script ... `s/(docker.io|quay.io|registry.access.redhat.com|registry.redhat.io)/artifactory.private.corp.com/g`

PR #216, opened 2019-09-04
Release Che 7.1.0  version of the plugin-registry

ibuziuk:
### What does this PR do?

Release Che 7.1.0  version of the plugin-registry https://github.com/eclipse/che/issues/14352:

- Bumping VERSION file
- Adding '7.1.0' versions for che-theia & machine-exec
- Setting 'latest' version for che-theia and machine-exec plugins to '7.1.0' (latest = latest release version)
- Setting '7.1.0' version for container runtime images of the che remote plugins - `find . -name '*.yaml' -exec sed -i -e 's/:next/:7.1.0/g' {} \;`

Please review in commit-by-commit manner


benoitf:
hello, AFAIK branch had by conventions `7.1.x` and tags were `7.1.0`
here it seems that branch and tags will be the same ?

ibuziuk:
@benoitf hi, we did the same for `7.0.0` release so I also created `7.1.0` the very same way. Currently, there are no tags for plugin / devfile registry at all. But I agree we do need  to standardize the release process since currently it is pretty messy

benoitf:
@ibuziuk https://github.com/eclipse/che-plugin-registry/tree/7.0.x (it's called `7.0.x` no ?)

v3/plugins/eclipse/che-theia/7.1.0/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+name: che-theia
+version: 7.1.0
+type: Che Editor
+displayName: theia-ide
+title: Eclipse Theia development version.
+description: Eclipse Theia, get the latest release each day.
+icon: https://raw.githubusercontent.com/theia-ide/theia/master/logo/theia-logo-no-text-black.svg?sanitize=true
+category: Editor
+repository: https://github.com/eclipse/che-theia
+firstPublicationDate: "2019-03-07"
```
benoitf: I would say that it's today the first publication date for che-theia 7.1.0

ibuziuk:
ok, I can rename the branch but it will not be consistent with https://github.com/eclipse/che-devfile-registry

v3/plugins/eclipse/che-theia/7.1.0/meta.yaml
ibuziuk: good point, but I believe we have not updated the `firstPublicationDate` before and it might be assumed that `firstPublicationDate` = firstPublicationDate of plugin in general, but I'm not sure

ibuziuk:
@benoitf changed the branch to `7.1.x` also done the same for  devfile-registry:
- branch `7.1.x` - https://github.com/eclipse/che-devfile-registry/tree/7.1.x
- tag `7.1.0` - https://github.com/eclipse/che-devfile-registry/tree/7.1.0

PR #217, opened 2019-09-04
Bumping version to 7.2.0-SNAPSHOT , adding '7.1.0' versions of che-theia and machine-exec and updating 'latest' version of those to '7.1.0'

ibuziuk:
### What does this PR do?

Bumping version to 7.2.0-SNAPSHOT , adding '7.1.0' versions of che-theia and machine-exec and updating 'latest' version of those to '7.1.0'

https://github.com/eclipse/che/issues/14352


benoitf:
shouldn't che-theia and machine-exec be with 'next' ?

benoitf:
ok, misread with latest, you can ignore

PR #221, opened 2019-09-10
[Proposal] Improve plugin icons

vitaliy-guliy:
### What does this PR do?

Sets proper icons for each plugin

Screenshot from 2019-09-10 15-37-55

Screenshot from 2019-09-10 15-38-24

The PR has to be finished, it's only first steps to demonstrate how it may look.
To test how it works, just add following JSON file as a custom plugin registry to your Che-Theia

```
https://raw.githubusercontent.com/vitaliy-guliy/che-plugin-registry/master/v3/plugins/plugins.json
```

Currently PNG images are not allowed to be used in the registry. We need to something with that because almost all the VS Code extensions use this format for icons.

Issue to allow PNG images https://github.com/eclipse/che/issues/14502
Issue for this improvement https://github.com/eclipse/che/issues/14510


benoitf:
what are copyrights on the images ?

vitaliy-guliy:
@benoitf we don't include the images in our repository as we don't include the third party .vsix binaries. Do we need to take care about that?

vitaliy-guliy:
@benoitf yep. I forgot to mention that in the PR description.
https://raw.githubusercontent.com/vitaliy-guliy/che-plugin-registry/master/v3/plugins/eclipse-che-logo.png
This logo I found on our sites and have been reworked it a bit.
I think we need to upload it somewhere to be public.

Let's say. I will create a dedicated PR wit this one image and place it in `v3/plugins` directory
https://github.com/eclipse/che-plugin-registry/tree/master/v3/plugins

After updating https://che-plugin-registry.openshift.io/ this image will be accessible by URI
https://che-plugin-registry.openshift.io/v3/plugins/eclipse-che-logo.png

And only then I will update the links to this image and update this PR. Is it ok?

mmorhun:
@vitaliy-guliy I would like to have at least separate folder for images.

vitaliy-guliy:
@benoitf @mmorhun 
`v3/plugins` for plugins
`v3/images` for images

amisevsk:
> After updating https://che-plugin-registry.openshift.io/ this image will be accessible by URI
https://che-plugin-registry.openshift.io/v3/plugins/eclipse-che-logo.png

We should not depend on che-plugin-registry.openshift.io since the primary priority there is supporting hosted Che. For now, no updates will be rolled out there until the next release.

In the future, new Che deploys will depend on deploying an instance of the registries, since we cannot guarantee version compatibility between the hosted registry and the version of Che you're installing.

amisevsk:
The official Che logo should be hosted somewhere on eclipse.org, IMO.

benoitf:
also in case of private networks/ airgap we may not have access to external URLs so all resources should be part of the registry. (che-plugin-registry.openshift.io or eclipse.org may not be reachable at all)

vitaliy-guliy:
I don't see any problem if the plugin metadata refers the icon on the same host.

In our case file 

https://che-plugin-registry.openshift.io/v3/plugins/eclipse/che-machine-exec-plugin/latest/meta.yaml

could refer on

https://che-plugin-registry.openshift.io/v3/images/eclipse-che-logo.png

> The official Che logo should be hosted somewhere on eclipse.org, IMO.

I don't mind. But also I don't see any issues to host the image on che-plugin-registry.openshift.io

benoitf:
@vitaliy-guliy yes in that case vsix files are stored in the registry as well

amisevsk:
> also in case of private networks/ airgap we may not have access to external URLs so all resources should be part of the registry. (che-plugin-registry.openshift.io or eclipse.org may not be reachable at all)

This is still not really possible to do; hosting the icon in the registry requires knowing the URL of the registry before building. I'm considering adding this to https://github.com/eclipse/che-plugin-registry/pull/211 but it's not clear that there's a good way to do this without changing the deploy process significantly.

amisevsk:
> I don't see any problem if the plugin metadata refers the icon on the same host.

Most deploys of Che should not refer to che.openshift.io at all. We can use your patch for hosted Che but any other deployment of Che will come with its own registries; these icons would no longer be on the same host.

vitaliy-guliy:
@benoitf then it's not a problem. You may use default icon for all the plugins. If this is a private network, that could be normal.
But anyway you have to rework/fix all the meta files in the plugin registry in that case.

vitaliy-guliy:
@amisevsk we are talking about default icon or for the icons for VS Code extensions?

vitaliy-guliy:
I still don't understand why it's an issue to refer on the VS Code extension icon having a direct reference on the VS Code binary.

amisevsk:
@vitaliy-guliy Sorry the conversation is getting a little muddled. I'm referring to 

> https://raw.githubusercontent.com/vitaliy-guliy/che-plugin-registry/master/v3/plugins/eclipse-che-logo.png
> This logo I found on our sites and have been reworked it a bit.
> I think we need to upload it somewhere to be public.

> Let's say. I will create a dedicated PR wit this one image and place it in v3/plugins directory
https://github.com/eclipse/che-plugin-registry/tree/master/v3/plugins

> After updating https://che-plugin-registry.openshift.io/ this image will be accessible by URI
https://che-plugin-registry.openshift.io/v3/plugins/eclipse-che-logo.png

> And only then I will update the links to this image and update this PR. Is it ok?

The logo cannot be hosted on https://che-plugin-registry.openshift.io, since this repo should not depend on that registry being in a certain state. If we want to host the new logo `.png`, it should be pushed to either https://github.com/eclipse/che/tree/master/dashboard/src/assets/branding or the eclipse.org site.

vitaliy-guliy:
@amisevsk
I see two options here.

The first one
> The logo cannot be hosted on https://che-plugin-registry.openshift.io, since this repo should not depend on that registry being in a certain state

There could be a relative reference. It will solve problem with referring on https://che-plugin-registry.openshift.io
Che-Theia knows the direct URI to meta.yaml. It can find the image by relative path.

The second one
> it should be pushed to either https://github.com/eclipse/che/tree/master/dashboard/src/assets/branding 

In that case what should be in meta.yaml and how we can calculate the icon URI?
Che-Theia is running in container and knows nothing about dashboard.

vitaliy-guliy:
> If Theia can handle relative references, then that would work.

We can implement that. It's not difficult.

> It would make implementing icon caching for #211 a lot easier as well

Icons are loaded by the browser. I believe it will cache everything.

> In this case, the meta.yaml would just refer to the raw file on github: e.g.
> https://raw.githubusercontent.com/eclipse/che/master/dashboard/src/assets/branding/loader.svg
> Che branding and icons should definitely be hosted somewhere in general.

Well. Then let's choose from options

- Che logo image is hosted on https://che-plugin-registry.openshift.io/v3/icons and accessible by relative URL
- Che logo image is uploaded to the GitHub and we just use the direct link to the image e.g. https://raw.githubusercontent.com/eclipse/che/master/dashboard/src/assets/branding/loader.svg

nickboldt:
For airgap `productization` we need icons cached in the plugin registry. Or else you'll have to fallback to ...
* no icon
* default icon
* something broken

So, +1 to cache images inside the plugin registry. 

If PR #211 ever gets merged we can reuse /resources/ folder for icons. If not and there's a compelling reason to keep 'em separated (to borrow a line from The Offspring) ... then that works too.

If the vsix files already contain an icon, could we not just use that? Sure, they might not be in the same location (root folder, extension/ folder) ...

image

But for cases where there's no visx file... like https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/eclipse/che-machine-exec-plugin/latest/meta.yaml ... then I guess we need an alternative approach.

nickboldt:
> what are copyrights on the images ?

They'd be covered by the same extension/LICENSE.txt in the .vsix archive. Eg., for ms-python, (c) Microsoft, and MIT License.

For fabric8-analytics, Apache 2.0 License.

image

So if it's safe to include the vsix in an offline/airgap plugin registry, the image is safe too.

nickboldt:
Is this still in progress? Needs rebasing and a correctly signed ECA on the commit if so.

nickboldt:
This is so old it still contains /latest/meta.yaml (from before Che 7.3) files so I'm going to close it as won't fix. Feel free to rebase and reopen if it's still something we want to do.

PR #222, opened 2019-09-10
to make it easier to run the check*.sh locally

nickboldt:
to make it easier to run the check*.sh scripts locally, use $(dirname "$0")/util.sh instead of just ./util.sh

Change-Id: I08ac3380ba9f2a8ea3489a4dc77befe052f0cf84

### DIFF
--- a/scripts/check_plugins_location.sh
+++ b/scripts/check_plugins_location.sh
@@ -15,7 +15,7 @@
 set -e
 
 # shellcheck source=./scripts/util.sh
-source ./util.sh
+source "$(dirname "$0")/util.sh"
 
 readarray -d '' metas < <(find "$1" -name 'meta.yaml' -print0)
 for meta in "${metas[@]}"; do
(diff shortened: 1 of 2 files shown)


benoitf:
FYI there is shellcheck error

scripts/check_plugins_location.sh
```diff
@@ -15,7 +15,7 @@
 set -e
 
 # shellcheck source=./scripts/util.sh
-source ./util.sh
+source $(dirname "$0")/util.sh
```
amisevsk: Looks like a job for good old `$(cd "$(dirname "$0")"; pwd)` to avoid relative path stuff.

nickboldt: Nope, shellcheck complains about missing quotes around the path.

nickboldt:
Can we add more CODE_OWNERS? We've got 3 approvals now and still can't merge. :(

image

amisevsk:
@nickboldt https://github.com/eclipse/che-plugin-registry/pull/223 :)

nickboldt:
Now I have a way to push stuff.  Ho-ho-ho :)

https://media2.giphy.com/media/oMK5z1GT0B87S/source.gif

PR #226, opened 2019-09-20
use vscode-python 2019.2.5558 instead of 2019.2.5433

nickboldt:
switch to vscode-python/releases/download/2019.2.5558 (in GH releases) vs. ms-python/vsextensions/python/2019.2.5433/vspackage (only in marketplace) (#14573)

Change-Id: I678ba2eb95ba2b2d1529b8d46d8b5aa0b9a10698

### DIFF
--- a/v3/plugins/ms-python/python/2019.2.5558/meta.yaml
+++ b/v3/plugins/ms-python/python/2019.2.5558/meta.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 publisher: ms-python
 name: python
-version: 2019.2.5433
+version: 2019.2.5558
 type: VS Code extension
 displayName: Python
 title: Python extension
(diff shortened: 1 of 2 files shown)


benoitf:
fYI link (#14573) is not clickable

nickboldt:
thx for the +1, now we just need a code owner review. :(

benoitf:
yes I'm not one

v3/plugins/ms-python/python/2019.2.5558/meta.yaml
```diff
@@ -16,4 +16,4 @@ spec:
       name: "vscode-python"
       memoryLimit: '512Mi'
   extensions:
-    - https://marketplace.visualstudio.com/_apis/public/gallery/publishers/ms-python/vsextensions/python/2019.2.5433/vspackage
+    - https://github.com/microsoft/vscode-python/releases/download/2019.2.5558/ms-python-release.vsix
```
amisevsk: Trailing newline

nickboldt: I think you mean no newline at end of line. Might be my github settings that converted from 'nix to non-'nix or something.

nickboldt: > Worth noting that these changes will break any workspaces using this registry and the old 2019.2.5433 version explicitly.

Yes, but do we want compliance, or compatibility? :)

PR #228, opened 2019-09-23
Add plugin deprecation info into registry index file

mmorhun:
Signed-off-by: Mykola Morhun <mmorhun@redhat.com>

### What does this PR do?
Adds `deprecate` section of plugin meta information into `index.json` file of the plugin registry.
Is needed for dashboard to be able to display and suggest auto migration. See https://github.com/eclipse/che/issues/13021


sleshchenko:
It would be useful to see examples of script work in both cases - for non-deprecated and for deprecated plugins.

sleshchenko:
Also please consider changing index.json items model in README.md https://github.com/eclipse/che-plugin-registry#get-index-list-of-all-plugins

mmorhun:
@sleshchenko I described it in the description, but yes, an example would be even better.
Updated description.

mmorhun:
As about readme, we have already [described](https://github.com/eclipse/che-plugin-registry#plugin-meta-yaml-structure) the `deprecate` section. It is just copied to index.json for such plugind, so no need in update.

sleshchenko:
> As about readme, we have already described the deprecate section. It is just copied to index.json for such plugind, so no need in update.

So, but README.md does not give any clue that `deprecate` is copied to index.json.
I still think that it would be better to paste example of deprecate section to index.json example.

mmorhun:
@sleshchenko updated

README.md
```diff
@@ -173,6 +173,34 @@ Response:
...
+    "displayName": "x lang support",
+    "version": "2019.08.20",
+    "type": "VS Code extension",
+    "name": "x-lang-ls",
+    "description": "Provides support for language x",
+    "publisher": "eclipse",
+    "links": {
+      "self": "/v3/plugins/eclipse/x-lang-ls/2019.08.20"
+    },
+    "deprecate": {
```
sleshchenko: just one minor comment - I'd put links after deprecate.

scripts/index.sh
```diff
@@ -24,5 +24,7 @@ yq -sS 'map(
     "\(.publisher)/\(.name)/\(.version)" as $id |
     {
         $id, displayName, version, type, name, description, publisher,
-        links: {self: "\($PLUGINS_DIR)/\($id)"}
-    }) | sort_by(.id)' "${metas[@]}" --arg "PLUGINS_DIR" "$PLUGINS_DIR"
+        links: {self: "\($PLUGINS_DIR)/\($id)"},
+        deprecate
+    } | if .deprecate == null then del(.deprecate) else . end) |
```
amisevsk: This can also be made a little cleaner I think:
```
yq -sS 'map(
    "\(.publisher)/\(.name)/\(.version)" as $id |
    {
        $id, displayName, version, type, name, description, publisher,
        links: {self: "\($PLUGINS_DIR)/\($id)"}
    } + if has("deprecate") then {deprecate} else null end ) | 
    sort_by(.id)' "${metas[@]}" --arg "PLUGINS_DIR" "$PLUGINS_DIR"
```

mmorhun: Thanks! This is better way of adding deprecate field in the index json file.

PR #229, opened 2019-09-24
Add dependency analytics extension v0.0.13 to plugin registry

arajkumar:
### What does this PR do?

Adds meta information for Dependency Analytics Che Plugin v0.0.13/latest

Dependent PR :
eclipse/che-theia#469

### DIFF
--- a/v3/plugins/redhat/dependency-analytics/0.0.13/meta.yaml
+++ b/v3/plugins/redhat/dependency-analytics/0.0.13/meta.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 publisher: redhat
 name: dependency-analytics
-version: latest
+version: 0.0.13
 type: VS Code extension
 displayName: Dependency Analytics
 title: Insights about your application dependencies


v3/plugins/redhat/dependency-analytics/0.0.13/meta.yaml
```diff
@@ -0,0 +1,18 @@
...
+name: dependency-analytics
+version: 0.0.13
+type: VS Code extension
+displayName: Dependency Analytics
+title: Insights about your application dependencies
+description: Insights about your application dependencies, Security, License compatibility and AI based guidance to choose appropriate dependencies for your application.
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/fabric8-analytics/fabric8-analytics-vscode-extension
+category: Other
+firstPublicationDate: '2018-10-03'
```
amisevsk: Please update `firstPublicationDate` to e.g. the date of this PR (`2019-09-24`) and do similar for the latest version.

v3/plugins/redhat/dependency-analytics/0.0.13/meta.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  containers:
+    - image: "docker.io/eclipse/che-remote-plugin-dependency-analytics-0.0.13:next"
+      memoryLimit: "512Mi"
```
amisevsk: Is `512Mi` really necessary for this plugin? Such a high default value means it cannot easily be used in many workspaces.

arajkumar: I thought `memoryLimit` caps the max RAM usage. Does it reserves the given amount of memory?

v3/plugins/redhat/dependency-analytics/0.0.13/meta.yaml
arajkumar: I can address this in the next iteration. Thanks.

v3/plugins/redhat/dependency-analytics/0.0.13/meta.yaml
amisevsk: On the OpenShift side, it is a cap (with request as the lower bound), but on clusters that aren't configured to overprovision memory, they still eat into quotas. 

E.g. on Hosted Che, the maximum memory limit for a Che workspace is `3Gi`

arajkumar: I tried reducing to 256Mi, it is failing on some big maven stacks like [che](https://github.com/eclipse/che). I see the following error,
```
Initial heap size set to a larger value than the maximum heap size
```

arajkumar: I tried to use the `requests` attribute, it didn't seem to be working.

```
  containers:
    - image: "docker.io/eclipse/che-remote-plugin-dependency-analytics-0.0.12:7.1.0"
      resources:
        requests:
          memory: "128Mi"
        limits:
          memory: "512Mi"
```

amisevsk: I don't think requests are taken into account at the moment.

nickboldt: If 256 is too small and 512 is too big, what about 384?

arajkumar: Memory consumption is based on the stack which we scan. Stacks like [che](https://github.com/eclipse/che) consumes almost 500MB while doing the scan. Anyways, `memoryLimit` can be overridden from the devfile., which gives more control to the users to reduce/increase the value based on the stack they scan. I'm fine with either 384 or 512.

PR #232, opened 2019-09-27
add --latest-only caching option

nickboldt:
Change-Id:...

add --latest-only caching option

Change-Id: Ie0690a1ab8f9e89989b658a05b890075635c92e6


nickboldt:
BIggest change in this set is moving from Dockerfile.rhel (uses alpine + `quay.io/openshiftio/rhel-base-httpd:latest`) to rhel.Dockerfile (uses images from RHCC: `ubi8-minimal` and `registry.access.redhat.com/rhscl/httpd-24-rhel7:2.4-104`)

Dockerfile
```diff
@@ -14,12 +17,16 @@ RUN apk add --no-cache py-pip jq bash wget && pip install yq jsonschema
...
-RUN ./set_plugin_dates.sh v3
-RUN ./check_plugins_viewer_mandatory_fields.sh v3
-RUN ./ensure_latest_exists.sh
-RUN ./index.sh v3 > /build/v3/plugins/index.json
-RUN chmod -R g+rwX /build
+
+# if only including the /latest/ plugins, apply this line to remove them from builder 
+# RUN rm -fr $(find /build/v3 -name 'meta.yaml' | grep -v "/latest/")
+
+RUN ./check_plugins_location.sh v3 && \
```
amisevsk: I'd rather not combine the `RUN` command when in the builder image -- this is one of the issues multi-stage builds were designed to address. 

It's harder to read when a failure occurs (rather than `./ensure_latest_exists.sh` failed, we get the whole long string failed).

Dockerfile
```diff
@@ -28,8 +35,10 @@ COPY --from=builder /build/v3 /var/www/html/v3
 
 # Offline build: cache .theia and .vsix files in registry itself and update metas
 FROM builder AS offline-builder
-RUN ./cache_artifacts.sh v3
-RUN chmod -R g+rwX /build
+
+# To only cache files from /latest/ folders, use ./cache_artifacts.sh v3 --latest-only 
+# and uncomment line above to remove files so they're not included in index.json -- RUN rm -fr $(find /build/v3 -name 'meta.yaml' | grep -v "/latest/")
```
amisevsk: Is this necessary? If an earlier step removes non-latest metas, then `cache_artifacts.sh` won't see the removed files and so only latest will be cached naturally.

README.md
```diff
@@ -21,10 +21,10 @@ Note that the Dockerfiles feature multi-stage build, so it requires Docker versi
 
 ### Offline and airgapped registry images
 
-It's possible to build an image for the plugin registry that includes all referenced extension artifacts (i.e. all `.theia` and `.vsix` archives). This is done using the `Dockerfile.caching` dockerfile during the build, which downloads artifacts and rewrites the devfiles to extensions cached in `v3/resources/`.
+It's possible to build an image for the plugin registry that includes all referenced extension artifacts (i.e. all `.theia` and `.vsix` archives). This is done using the same `Dockerfile`, but performs additional steps to download artifacts and rewrite the devfiles to use files cached in `v3/resources/`.
```
amisevsk: Thanks for catching this, I forgot to update in my last PR.

rhel.Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+# COPY root-local.tgz /tmp/root-local.tgz
+
+# NOTE: uncomment for local build. Must also set full registry path in FROM to registry.redhat.io or registry.access.redhat.com
+# enable rhel 7 or 8 content sets (from Brew) to resolve jq as rpm
+COPY content_sets_epel7.repo /etc/yum.repos.d/
+
+RUN microdnf install -y findutils bash wget yum gzip tar jq python3-six python3-pip && microdnf -y clean all && \
+    # install yq (depends on jq and pyyaml - if jq and pyyaml not already installed, this will try to compile it)
+    if [[ -f /tmp/root-local.tgz ]]; then mkdir -p /root/.local && tar xf /tmp/root-local.tgz -C /root/.local/ && rm -fr /tmp/root-local.tgz; fi && \
+    /usr/bin/pip3.6 install --user yq jsonschema && \
```
amisevsk: I believe that if you don't use the `--user` install, you don't need the next "find-and-copy-into-$PATH" step.

rhel.Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+USER 0
+
+# BEGIN these steps might not be required
+RUN sed -i /etc/httpd/conf/httpd.conf \
+    -e "s,Listen 80,Listen 8080," \
+    -e "s,logs/error_log,/dev/stderr," \
+    -e "s,logs/access_log,/dev/stdout," \
+    -e "s,AllowOverride None,AllowOverride All," && \
+    chmod a+rwX /etc/httpd/conf /run/httpd
+EXPOSE 80
```
amisevsk: Is it necessary to expose port 80? We're rewriting the conf to listen on 8080.

AFAIK binding to ports <1000 will fail on OpenShift as it requires root.

scripts/cache_artifacts.sh
```diff
@@ -14,7 +14,11 @@
 
 set -e
 
-readarray -d '' metas < <(find "$1" -name 'meta.yaml' -print0)
+if [[ $2 == "--latest-only" ]]; then
```
amisevsk: This shouldn't be necessary if we clean non-latest metas from the filesystem in the builder image.

rhel.Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+#    cd /tmp/root-local && sudo tar cvzf root-local.tgz lib/ bin/ && cd - && mv -f /tmp/root-local/root-local.tgz . && sudo rm -fr /tmp/root-local/
+# 3. then add it to dist-git so it's part of this repo
+#    rhpkg new-sources root-local.tgz 
+
+# built in Brew, use tarball in lookaside cache
+# COPY root-local.tgz /tmp/root-local.tgz
+
+# NOTE: uncomment for local build. Must also set full registry path in FROM to registry.redhat.io or registry.access.redhat.com
+# enable rhel 7 or 8 content sets (from Brew) to resolve jq as rpm
+COPY content_sets_epel7.repo /etc/yum.repos.d/
```
amisevsk: WDYT about implementing this as an echo command? It's a weirder approach but also avoids adding the `content_sets_epel7.repo` file to the repo, which I think could be confusing for someone not familiar with rhel.

Alternatively, we could move dockerfiles and associated files into a `/build` directory in the repository, to make it clearer what belongs where.

benoitf:
why `rhel.Dockerfile` instead of `Dockerfile.rhel` ? it's hard to find it also it's not consistent with eclipse-che dockerfiles https://github.com/eclipse/che/tree/master/dockerfiles/che

rhel.Dockerfile
benoitf: yes sounds weird to have these repo files in root directory

rhel.Dockerfile
benoitf: yes we don't care on which port there is a listener

nickboldt:
> why rhel.Dockerfile instead of Dockerfile.rhel

the `Dockerfile` file extension is registered automatically in VSCode. The `Dockerfile.*` is not. This is why these files get autocompletion and linting:

- `Dockerfile`
- `.Dockerfile`
- `rhel.Dockerfile`
 
And this does not:

- `Dockerfile.rhel`

So to avoid bugs in the rhel version (which people are NOT using in upstream anyway, and aren't being run in CI) I'd rather at least make it easier to edit them :)

rhel.Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+      io.k8s.description="$DESCRIPTION" \
+      io.k8s.display-name="$DESCRIPTION" \
+      io.openshift.tags="$PRODNAME,$COMPNAME" \
+      com.redhat.component="$PRODNAME-$COMPNAME-container" \
+      name="$PRODNAME/$COMPNAME" \
+      version="2.0" \
+      license="EPLv2" \
+      maintainer="Nick Boldt <nboldt@redhat.com>" \
+      io.openshift.expose-services="" \
+      usage=""
```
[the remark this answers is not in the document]
nickboldt: Consistency with upstream, like we have in https://github.com/eclipse/che-operator/blob/master/Dockerfile

Doesn't break anything in upstream. Just makes up->down sync easier.

rhel.Dockerfile
nickboldt: +1 to use build/dockerfiles - consistency with https://github.com/eclipse/che-devfile-registry/tree/master/build/dockerfiles would be nice. I can move 'em if we agree.

As to the `echo > file` vs. `COPY` um... again, more consistent w/ up/downstream. In Brew this file isn't used (its info is injected by the Brew system). But to make these builds work locally outside Brew, the repo info is required. We could embed the information into a long echo line but OTOH if you concern is that people don't know stuff about rhel, then surely this would help them learn stuff, rather than hiding the details in the rhel.Dockerfile ?

scripts/cache_artifacts.sh
nickboldt: True, but this allows another option. Maybe you want to keep all the meta.yaml files but only CACHE the OFFLINE versions for the latest ones.

rhel.Dockerfile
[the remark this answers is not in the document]
benoitf: if it's in this codebase, then there is no single MAINTAINER, no com.redhat.component and no bloatware labels.

rhel.Dockerfile
nickboldt: This was in the existing [Dockerfile.rhel](https://github.com/eclipse/che-plugin-registry/blob/master/Dockerfile.rhel#L32), but if it's not needed, cool. Interestingly there's nothing in the existing `registry.centos.org/centos/httpd-24-centos7`-based Dockerfile that exposes any ports... 80, 8080, 443, 8443, etc. 

So... is it exposed by default in the upstream httpd-24-centos7 image?

rhel.Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+
+# To only cache files from /latest/ folders, use ./cache_artifacts.sh v3 --latest-only 
+# and uncomment line above to remove files so they're not included in index.json -- RUN rm -fr $(find /build/v3 -name 'meta.yaml' | grep -v "/latest/")
+RUN ./cache_artifacts.sh v3 && chmod -R g+rwX /build
+
+FROM registry AS offline-registry
+USER 0
+
+# Offline registry: copy updated meta.yamls and cached extensions
+ENV SUMMARY="Red Hat CodeReady Workspaces plugin registry container" \
```
benoitf: it's not codeready workspaces there.

rhel.Dockerfile
nickboldt: IIRC pip doesn't like to run as root so the --user flag was required here. Prove me wrong?

Dockerfile
nickboldt: yes, `remove files so they're not included in index.json` is not the same as choosing to cache or not cache them locally.

This will delete files; the cache_artifacts does not delete stuff, it only caches/creates stuff.

Dockerfile
nickboldt: well, true, but it also runs faster and creates fewer layers.

amisevsk: Fair enough, it doesn't really matter to me.

Dockerfile
amisevsk: But `cache_artifacts` works by reading the extensions from the metas -- if the metas no longer exist it can't read them.

rhel.Dockerfile
amisevsk: `docker inspect registry.centos.org/centos/httpd-24-centos7` gives me posts 8080 and 8443 as exposed. Same thing for `registry.access.redhat.com/rhscl/httpd-24-rhel7:2.4-104`.

scripts/cache_artifacts.sh
amisevsk: Fair enough.

rhel.Dockerfile
amisevsk: +1 on moving to `/build/dockerfiles`, and your explanation makes sense. Adding the content-sets file makes more sense to me if it's in a dockerfiles folder with the dockerfile.

rhel.Dockerfile
amisevsk: `pip` doesn't like to run as root in general since you can break stuff, but I don't know that the user flag helps a whole lot when your user is root (I still get the 
> WARNING: Running pip install with root privileges is generally not a good idea. Try `pip3.6 install --user` instead.

warning with the flag).

The dockerfile
```dockerfile
FROM registry.access.redhat.com/ubi8-minimal:8.0-213
USER 0

COPY content_sets_epel7.repo /etc/yum.repos.d/
RUN microdnf install -y findutils jq python3-six python3-pip && microdnf -y clean all

RUN /usr/bin/pip3.6 install yq

RUN echo '{"test": "hello"}' | yq '.'
```
works for me.

nickboldt: Yes but YOU have access to the public internet. Brew does not. So I install from the `root-local.tgz` instead of fetching from `https://files.pythonhosted.org/packages` for the whl and tar.gz files for dependencies. That's why my approach is more convoluted (but can fall back to fetching from the internet if the tarball isn't present).

Perhaps I should change the logic such that it omits the `--user` flag and all the `cp` commands if the tarball isn't found?

Dockerfile
nickboldt: Exactly. So we have to delete them if we want cache_artifacts to not see and therefore cache them. It's three steps:

1. delete meta.yaml files we don't want to include in index.json
2. generate index.json (excluding the deleted files!) w/ `index.sh`
3. cache only the /latest/ meta.yaml files' vsix/theia files with `cache_artifacts.sh`

I considered adding the delete functionality to the cache_artifacts script but if we do that we'd have to call the `index.sh` script again so we don't point at meta.yaml files that no longer exist.

Alternatively we could add a -latest-only flag to the index.sh script ... if you prefer that?

rhel.Dockerfile
nickboldt: OK, I've updated the logic so that it only installs to the /root/.local folder if BOOTSTRAP=true or there's a tarball copied into the container.

rhel.Dockerfile
nickboldt: Port 80 removed.

rhel.Dockerfile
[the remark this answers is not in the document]
nickboldt: Translation: if it's in an eclipse/ repo, you don't want RH metadata. So I have to maintain a forked version in https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-plugin-registry/Dockerfile :(

rhel.Dockerfile
nickboldt: can you just ignore it? it's not like you're using the rhel.Dockerfile to build che-plugin-registry, since you're using the alpine one :)

rhel.Dockerfile
nickboldt: Here's the thing though... if I move the `*Dockerfile*` files to /build/dockerfiles, I also have to move ... `README.md`, `v3/` and `scripts/` since they have to be local to the Dockerfile context.

Which means the root folder then only has:

```
build/
build.sh
cico_build_ci.sh
cico_build_nightly.sh
cico_build_release.sh
cico_functions.sh
CONTRIBUTE.md
kubernetes/
LICENSE
openshift/
VERSION
```

I mean if you think having build-related files in the root is wrong, then maybe we need to move all the cico* files and build.sh too?

rhel.Dockerfile
benoitf: No, I can't ;-)

rhel.Dockerfile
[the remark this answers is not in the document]
benoitf: Yes, that's correct that eclipse projects are not direct redhat products. I suppose we can use dockerarg or template there to invoke build script differently and not maintain forked dockerfile

rhel.Dockerfile
nickboldt: So I need these labels and envs in a separate file downstream, or applied via a transform/merge process. Fun. OK.

rhel.Dockerfile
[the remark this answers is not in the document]
nickboldt: I have removed the harmless metadata.

nickboldt:
two remaining questions are:

• about the .repo file (and reorging the gh repo itself to nest stuff under build/dockerfiles/ - which seems like a low-value change) and 
• the question about a simple `RUN rm -f ...` vs. adding a `--latest-only` flag to the index.sh script so it can ignore non-latest content when enerating index.json

Dockerfile
amisevsk: If there's a use-case for caching only latest plugins but keeping all metas available (so you can choose latest and it's cached, but if you chose any other version it has to download), then a `--latest-only` option could be useful. Honestly, though, that seems too complicated for so little gain (except your image would be smaller, I suppose). My preferred approach would be something like

- We support 2x2 options:
    - You can build offline or not (already implemented for the most part)
    - You can include all versions or only latest. If only latest, no other metas are stored in the registry.

The minimal changes to support this would be to add a way to delete non-latest metas in the builder phase, then everything would follow naturally without additional changes.

rhel.Dockerfile
amisevsk: Couldn't we specify the same context but put the dockerfile in a different dir?

```bash
docker build -t <image> -f ./build/dockerfiles/Dockerfile .
```

The context is still the root of the repo there.

amisevsk: This is how we do it in the devfile registry: https://github.com/eclipse/che-devfile-registry/blob/master/build.sh#L20

Dockerfile
nickboldt: aaand I already have that.

```
RUN rm -fr $(find /build/v3 -name 'meta.yaml' | grep -v "/latest/")
```

in: 

* Dockerfile line 22 (optional, commented out): https://github.com/eclipse/che-plugin-registry/pull/232/files#diff-3254677a7917c6c01f55212f86c57fbfR22
* rhel.Dockerfile line 75 (optional, commented out): https://github.com/eclipse/che-plugin-registry/pull/232/files#diff-75aafc399136415e2e24dad895058632R75

rhel.Dockerfile
nickboldt: yeah, but can we handle that in a second PR? +1 as is if you like the current version and I'll propose a reorg change later so that I don't mess up the cico scripts too

benoitf:
about Docker.rhel and rhel.Docker pattern I think we should not be driven by a VS Code extension.

also you've a way to make it working
https://github.com/microsoft/vscode-docker/issues/192

rhel.Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+################# 
+
+# Build registry, copying meta.yamls and index.json from builder
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/rhscl/httpd-24-rhel7
+FROM registry.access.redhat.com/rhscl/httpd-24-rhel7:2.4-104 AS registry
+USER 0
+
+# BEGIN these steps might not be required
+RUN sed -i /etc/httpd/conf/httpd.conf \
+    -e "s,Listen 80,Listen 8080," \
```
amisevsk: `s,Listen 80,Listen 8080,` is not required as the default is `Listen 0.0.0.0:8080` (so there aren't matches), but the other lines do make changes (still don't know if they're necessary).

rhel.Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+    -e "s,logs/error_log,/dev/stderr," \
+    -e "s,logs/access_log,/dev/stdout," \
+    -e "s,AllowOverride None,AllowOverride All," && \
+    chmod a+rwX /etc/httpd/conf /run/httpd
```
amisevsk: I believe these dirs are already access mode `777`.

rhel.Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+WORKDIR /var/www/html
+ENTRYPOINT ["/usr/local/bin/uid_entrypoint.sh", "/usr/local/bin/entrypoint.sh"]
+
+# Offline build: cache .theia and .vsix files in registry itself and update metas
+# multiple temp stages does not work in Brew
+FROM builder AS offline-builder
+
+# To only cache files from /latest/ folders, use ./cache_artifacts.sh v3 --latest-only 
+# and uncomment line above to remove files so they're not included in index.json -- RUN rm -fr $(find /build/v3 -name 'meta.yaml' | grep -v "/latest/" | grep -o ".*/")
+# RUN ./cache_artifacts.sh v3 && chmod -R g+rwX /build
```
amisevsk: I think this is mistakenly commented out, there's no commands in `offline-builder` currently.

scripts/uid_entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "${USER_NAME:-jboss}:x:$(id -u):0:${USER_NAME:-jboss} user:${HOME}:/sbin/nologin" >> /etc/passwd
+  fi
+fi
+exec "$@"
```
amisevsk: Missing newline at end :)

Dockerfile
```diff
@@ -28,8 +35,10 @@ COPY --from=builder /build/v3 /var/www/html/v3
 
 # Offline build: cache .theia and .vsix files in registry itself and update metas
 FROM builder AS offline-builder
-RUN ./cache_artifacts.sh v3
-RUN chmod -R g+rwX /build
+
+# To only cache files from /latest/ folders, use ./cache_artifacts.sh v3 --latest-only 
+# and uncomment line above to remove files so they're not included in index.json -- RUN rm -fr $(find /build/v3 -name 'meta.yaml' | grep -v "/latest/" | grep -o ".*/")
```
amisevsk: For the non-rhel dockerfile, this could be implemented as an ARG (`ONLY_LATEST=true`) to avoid having to edit the dockerfile for specific builds.

rhel.Dockerfile
nickboldt: So you want me to remove this? or add -c to see what actually happens, if anything?

rhel.Dockerfile
```diff
@@ -0,0 +1,129 @@
+#
```
benoitf: Still a mistake to me to have rhel.Dockerfile instead of Dockerfile.rhel. It spread related files to different locations

content_sets_epel7.repo
```diff
@@ -0,0 +1,5 @@
+[epel-7]
```
benoitf: this file should be suffixes/prefixed by rhel to know it's only for rhel

scripts/entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -x
```
benoitf: It i missing copyright headers

scripts/uid_entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+if ! whoami &> /dev/null; then
```
benoitf: Missing copyright headers

scripts/uid_entrypoint.sh
```diff
@@ -0,0 +1,7 @@
...
+  if [ -w /etc/passwd ]; then
+    echo "${USER_NAME:-jboss}:x:$(id -u):0:${USER_NAME:-jboss} user:${HOME}:/sbin/nologin" >> /etc/passwd
```
benoitf: Why jboss ?

PR #234, opened 2019-10-01
add code-server to registry

sr229:
## What does this PR do?

This PR includes a regular plugin and a OpenShift-specific plugin of [code-server](https://github.com/cdr/code-server).

Resolves eclipse/che#12854


sr229:
Forgot to sign ECA, will do rn

benoitf:
@sr229 also it doesn't build
 ---> Running in d4fa1dff8af4
!!! Location mismatch in plugin 'cdr/eclipseide/2.1523-vsc1.38.1':
!!!   Expected location: 'v3/plugins/cdr/eclipseide/2.1523-vsc1.38.1/meta.yaml'
!!!   Actual location: 'v3/plugins/cdr/code-server/2.1523-vsc1.38.1/meta.yaml' 
The command '/bin/sh -c ./check_plugins_location.sh v3' returned a non-zero code: 1

sr229:
> @sr229 also it doesn't build
> ---> Running in d4fa1dff8af4
> !!! Location mismatch in plugin 'cdr/eclipseide/2.1523-vsc1.38.1':
> !!! Expected location: 'v3/plugins/cdr/eclipseide/2.1523-vsc1.38.1/meta.yaml'
> !!! Actual location: 'v3/plugins/cdr/code-server/2.1523-vsc1.38.1/meta.yaml'
> The command '/bin/sh -c ./check_plugins_location.sh v3' returned a non-zero code: 1

I figured out the error as soon as you replied haha, should be fixed now I think.

sr229:
Somehow CircleCI is reporting its still building but according to CircleCI's webapp, it just finished successfully, probably a lag on their endpoint.

image

v3/plugins/cdr/code-server/2.1523-vsc1.38.1-openshift/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+      attributes:
+        protocol: http
+        type: ide
+  containers:
+   - name: code-server
+     image: "docker.io/chinodesuuu/coder:v2-openshift"
+     mountSources: true
+     ports:
+         - exposedPort: 9000
+     memoryLimit: "1024M"
```
amisevsk: Please add newline at end of file

v3/plugins/cdr/code-server/2.1523-vsc1.38.1/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+      attributes:
+        protocol: http
+        type: ide
+  containers:
+   - name: code-server
+     image: "docker.io/chinodesuuu/coder:che-v2"
+     mountSources: true
+     ports:
+         - exposedPort: 9000
+     memoryLimit: "1024M"
```
amisevsk: Please add newline at end of file.

v3/plugins/cdr/code-server/latest
```diff
@@ -0,0 +1 @@
+2.1523-vsc1.38.1
```
amisevsk: The way latest currently works is that you need folders with a copy of the latest meta.yaml (I know, it's a hack-y solution). 

For selecting `latest` in the Che UI to work, this should be
- File: `v3/plugins/cdr/code-server/latest/meta.yaml`
- Content: Copy of `/v3/plugins/cdr/code-server/2.1523-vsc1.38.1/meta.yaml`, with `.version=latest`.

v3/plugins/cdr/code-server/2.1523-vsc1.38.1-openshift/meta.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: v2
+publisher: cdr
+name: code-server
+version: 2.1523-vsc1.38.1-openshift
```
amisevsk: Could you mention the differences in the `-openshift` version? I foresee some confusion in using this plugin on OpenShift (especially as latest looks like it won't be OpenShift-compatible).

sr229: OpenShift version just adds OpenShift-specific user ID fixes since OSO has a known security feature that scrambles UIDs, which breaks a lot of containers.

sr229: Docker containers are built [here](https://github.com/sr229/code-server-openshift) which should give you an idea why openshift is a bit more different.

amisevsk: In that case, to smooth user experience, what do you think about using the OpenShift-compatible image in both versions (k8s and OpenShift)? As far as I know, the usual OpenShift fixes aren't incompatible with Kubernetes (and it's possible to configure your kube cluster to do the same arbitrary UID stuff IIRC).

sr229: Sounds like a plan 👍

v3/plugins/cdr/code-server/2.1523-vsc1.38.1-openshift/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+firstPublicationDate: "2019-09-08"
+spec:
+  endpoints:
+   -  name: "code-server"
+      public: true
+      targetPort: 9000
+      attributes:
+        protocol: http
+        type: ide
+  containers:
```
amisevsk: One OpenShift-specific security fix that is missing: containers cannot be created on some clusters if they mount a volume that isn't also mounted as a k8s/OpenShift PVC:
```
Error: Error response from daemon: create 84eb3536d3dc596a7f79926e3b1f922adc77a19e2c5eec2e0b038b7437e7dd58: error while creating volume path '/var/lib/docker/volumes/84eb3536d3dc596a7f79926e3b1f922adc77a19e2c5eec2e0b038b7437e7dd58/_data': mkdir /var/lib/docker/volumes/84eb3536d3dc596a7f79926e3b1f922adc77a19e2c5eec2e0b038b7437e7dd58: permission denied
```

To resolve this, you would need to mount any volumes from the dockerimage as PVCs. The good news is we recently added support for this -- you can specify
```
containers:
    volumes:           
        - mountPath:
          name:
          ephemeral: # boolean; if true volume will be ephemeral, otherwise volume will be persisted
```
Looking at your current Dockerfile, you would need to add mounts for `/home/coder/projects` and `/home/coder/certs` here.

Alternatively, since the openshift-specific image is built using its own dockerfile, I would consider removing the `VOLUME` directives, since you can always add volumes in your k8s/OpenShift yaml.

sr229: Will do that right now, thanks for the feedback.

sr229:
> I tested out the OpenShift version (on OpenShift) and it starts up vscode in my browser, really cool.
> 
> The editor doesn't work with the Theia-specific machinery we have currently (clone project, plugins, etc), but as a first contribution it seems to work.
> 
> @sr229 One potential pain point: is it possible to configure code-server to start opening a specific folder? Currently, the only volume that would be persisted is mounted at `/projects`, the default is to open `/home/coder` or something similar; if a user checks out a git project there, it will disappear when the workspace shuts down.
> 
> Also, how does adding plugins work with code-server? I'm assuming they will also be wiped with the workspace shutting down.

Yes, its possible to point it to `/projects`, you can just invoke it like so
```
/opt/code-server --port=9000 /projects
```
And yes, unfortunately, the plugins won't persist if the IDE goes down, that's something I think I can't solve since that requires some volume trickery which I believe others are willing to do.

amisevsk:
> And yes, unfortunately, the plugins won't persist if the IDE goes down, that's something I think I can't solve since that requires some volume trickery which I believe others are willing to do.

We can play with this later, it might be as simple as adding a volume for the plugin but I'd need to check.

sr229:
@amisevsk I pushed a new Docker tag so its easy to version them properly. You can find the `che-v2` image source [here](https://github.com/sr229/code-server-openshift/blob/master/che-v2.dockerfile).

sr229:
Went with the route of persisting the entire user data dir for code-server. It should be standard VSCode directory, but anyone can refine those PVCs soon.

v3/plugins/cdr/code-server/2.1523-vsc1.38.1/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  endpoints:
+   -  name: "code-server"
+      public: true
+      targetPort: 9000
+      attributes:
+        protocol: http
+        type: ide
+  containers:
+   - name: code-server
+     image: "docker.io/chinodesuuu/coder:2.1523-vsc1.38.1-che"
```
l0rd: @sr229 Thanks for the PR! We try to avoid linking personal dockerhub repositories. Would you be able to push the image to `docker.io/codercom` repo instead?

sr229: For now that isn't possible, but I'll ask if I can push to the Docker namespace soon.

l0rd: ack

sr229:
I won't be able to change it to a `codercom/` docker container, but I'll be doing a follow-up PR to change that later on if that's fine for everyone.

l0rd:
LGTM but need to test it to approve it. @amisevsk if you have tested it and you are ok please proceed with merge.

l0rd:
Thanks @sr229 for this PR. Looking forward for next ones to improve coder support.

PR #237, opened 2019-10-01
add ARG LATEST_ONLY for better branching;...

nickboldt:
add ARG LATEST_ONLY for better branching; apply fixes in rhel.Dockerfile to Dockefile too

Change-Id: I19ab4e8b34b96deeae507e3795e6964fa0e6dbfb

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,10 @@
 
 # Builder: check meta.yamls and create index.json
 FROM alpine:3.10 AS builder
+USER 0
+
+ARG LATEST_ONLY=false
+ENV LATEST_ONLY=${LATEST_ONLY}
 RUN apk add --no-cache py-pip jq bash wget && pip install yq jsonschema
 
 COPY ./scripts/*.sh ./scripts/meta.yaml.schema /build/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -30,17 +36,21 @@ RUN ./check_plugins_location.sh v3 && \
...
 # Offline build: cache .theia and .vsix files in registry itself and update metas
 FROM builder AS offline-builder
 
-# To only cache files from /latest/ folders, use ./cache_artifacts.sh v3 --latest-only 
-# and uncomment line above to remove files so they're not included in index.json -- RUN rm -fr $(find /build/v3 -name 'meta.yaml' | grep -v "/latest/" | grep -o ".*/")
 RUN ./cache_artifacts.sh v3 && chmod -R g+rwX /build
 
 # Offline registry: copy updated meta.yamls and cached extensions
 FROM registry AS offline-registry
+USER 0
```
l0rd: This root user. Are you doing for a particular reason or is it just a copy/paste to be consistent with RHEL image? Do you know why this is needed in RHEL image?

nickboldt: we're already running as root, I'm just adding it there to be explicit. But it's not required.

PR #238, opened 2019-10-03
Add OpenShift Connector Che Plugin 0.1.0

azatsarynnyy:

### What does this PR do?
- adds descriptor file for OpenShift Connector Che Plugin 0.1.0.
- points latest version to 0.1.0

Should be merged once `eclipse/che-remote-plugin-openshift-connector-0.1.0:next` image is published.
See https://github.com/eclipse/che-theia/pull/465

### DIFF
--- a/v3/plugins/redhat/vscode-openshift-connector/latest/meta.yaml
+++ b/v3/plugins/redhat/vscode-openshift-connector/latest/meta.yaml
@@ -9,8 +9,7 @@ description: Interacting with Red Hat OpenShift clusters and providing a streaml
 icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
 repository: https://github.com/eclipse/che-plugin-registry/tree/master/v3/plugins/redhat/vscode-openshift-connector
 category: Other
-firstPublicationDate: "2019-03-11"
-latestUpdateDate: "2019-10-02"
+firstPublicationDate: "2019-10-08"
 spec:
   containers:
     - image: "docker.io/eclipse/che-remote-plugin-openshift-connector-0.1.0:next"
--- a/v3/plugins/redhat/vscode-openshift-connector/0.1.0/meta.yaml
+++ b/v3/plugins/redhat/vscode-openshift-connector/0.1.0/meta.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 publisher: redhat
 name: vscode-openshift-connector
-version: latest
+version: 0.1.0
 type: VS Code extension
 displayName: OpenShift Connector
 title: OpenShift Connector
--- a/v3/plugins/redhat/vscode-openshift-connector/0.1.0/meta.yaml
+++ b/v3/plugins/redhat/vscode-openshift-connector/0.1.0/meta.yaml
@@ -9,8 +9,7 @@ description: Interacting with Red Hat OpenShift clusters and providing a streaml
 icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
 repository: https://github.com/eclipse/che-plugin-registry/tree/master/v3/plugins/redhat/vscode-openshift-connector
 category: Other
-firstPublicationDate: "2019-03-11"
-latestUpdateDate: "2019-10-02"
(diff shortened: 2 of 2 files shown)


v3/plugins/redhat/vscode-openshift-connector/latest/meta.yaml
```diff
@@ -5,16 +5,16 @@ version: latest
...
-description: Interacting with Red Hat OpenShift clusters and providing a streamlined
-  developer experience using Eclipse Che
-icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
-repository: https://github.com/redhat-developer/vscode-openshift-tools
+description: Interacting with Red Hat OpenShift clusters and providing a streamlined developer experience using Eclipse Che
+icon: https://github.com/redhat-developer/vscode-openshift-tools/raw/master/images/openshift_extension.png
+repository: https://github.com/eclipse/che-plugin-registry/tree/master/v3/plugins/redhat/vscode-openshift-connector
 category: Other
-firstPublicationDate: '2019-04-19'
+firstPublicationDate: "2019-03-11"
```
amisevsk: Kind of funny to use an earlier date for the new latest version, since `0.1.0` is just now being released, does it make sense to use the date of the PR here instead?

azatsarynnyy: If I understand correctly, `firstPublicationDate` refers to the 1'st published version of the plugin and it shouldn't be changed in all future versions.
So, if the first published version of the plugin was [`0.0.17` which was published on 2019-03-11](https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/vscode-openshift-connector/0.0.17/meta.yaml#L12) that date should be in all the subsequent meta.yaml files.
And `latestUpdateDate` is the date when the latest version has been published.
@amisevsk  am I understand the meaning of these fields correctly?

azatsarynnyy:
> we still require .svg icons
@amisevsk thanks you for pointing that out! I've updated it

v3/plugins/redhat/vscode-openshift-connector/latest/meta.yaml
amisevsk: I'm not sure the intended spec for `firstPublicationDate`, but as it stands it seems to be updated on every release. 
```
$ yq '.firstPublicationDate' ./v3/plugins/redhat/java/**/meta.yaml
"2019-02-20"
"2019-04-25"
"2019-05-27"
"2019-06-18"
"2019-06-18"
```
Your interpretation does make more sense to me, though it does not seem to be the current approach. One issue with the current implementation is that no plugins have a `latestUpdateDate` -- it's set to todays date [during the build](https://github.com/eclipse/che-plugin-registry/blob/master/scripts/set_plugin_dates.sh). As a result, the helper script for setting up latest dirs also relies on each version having its [own `firstPublishedDate`](https://github.com/eclipse/che-plugin-registry/blob/b111e251c047fe7a960a32e1656b5b60f930e85a/scripts/setup_latest_dirs.sh#L30).

azatsarynnyy: @amisevsk thank you for the clarification! I've removed `latestUpdateDate` and set `firstPublicationDate` with current date.

amisevsk:
I'm looking into the CI failure, that should not be happening.

amisevsk:
https://github.com/eclipse/che-plugin-registry/pull/246 should fix CI

azatsarynnyy:
Looks like it reaches rate limit on marketplace.visualstudio.com

amisevsk:
@azatsarynnyy Please rebase on master to pick up the CI fix.

azatsarynnyy:
@amisevsk CI is green now.
Thank you!

v3/plugins/redhat/vscode-openshift-connector/latest/meta.yaml
amisevsk: :+1: we can potentially rework it later.

PR #239, opened 2019-10-04
Update java support

svor:
### What does this PR do?
- Adds java plugin v0.50.0
- Updates latest java plugin
- Updates java-debug plugin to v0.20.0
- Adds `/home/theia/.m2` volume for java plugin

Related issue:
https://github.com/eclipse/che/issues/14649
https://github.com/eclipse/che/issues/13603


amisevsk:
One feature that I don't know was intentional but is still very nice is that the mounted `.m2` repo is shared between the plugin container and the main java container (assuming the main container also mounts `.m2`)

PR #242, opened 2019-10-05
Use string value at `env.value` instead of boolean or integer.

monaka:
Signed-off-by: Masaki Muranaka <monaka@monami-ya.com>

### What does this PR do?

Fixes to use string value.

All values in `*.env.value` should be typed to string.
Since it is obvious that runtimes treat it as string.
It's not good to use implicit type cast, IMO.


monaka:
@amisevsk Thank you for merging this!

PR #252, opened 2019-10-11
fix brew bootstrap instructions

nickboldt:
Change-Id:...

fix brew bootstrap instructions

Change-Id: Ica1a7d316487682229e06ea02a0517455416ebe7

### DIFF
--- a/build/dockerfiles/rhel.Dockerfile
+++ b/build/dockerfiles/rhel.Dockerfile
@@ -20,10 +20,13 @@ USER 0
 ################# 
 
 ARG BOOTSTRAP=false
+ENV BOOTSTRAP=${BOOTSTRAP}
 ARG LATEST_ONLY=false
+ENV LATEST_ONLY=${LATEST_ONLY}
 
 # to get all the python deps pre-fetched so we can build in Brew:
 # 1. extract files in the container to your local filesystem
+#    find v3 -type f -exec dos2unix {} \;
 #    CONTAINERNAME="pluginregistrybuilder" && docker build -t ${CONTAINERNAME} . --target=builder --no-cache --squash --build-arg BOOTSTRAP=true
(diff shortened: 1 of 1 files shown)


build/dockerfiles/rhel.Dockerfile
```diff
@@ -20,10 +20,13 @@ USER 0
 ################# 
 
 ARG BOOTSTRAP=false
+ENV BOOTSTRAP=${BOOTSTRAP}
 ARG LATEST_ONLY=false
+ENV LATEST_ONLY=${LATEST_ONLY}
```
amisevsk: You shouldn't need env vars here, the ARGs are available during the docker build.

build/dockerfiles/rhel.Dockerfile
```diff
@@ -122,16 +125,21 @@ FROM builder AS offline-builder
...
 
 # 2. then add it to dist-git so it's part of this repo
 #    rhpkg new-sources root-local.tgz v3.tgz
-RUN if [ ! -f /tmp/v3.tgz ] || [ ${BOOTSTRAP} == "true" ]; then \
-      ./cache_artifacts.sh v3 && chmod -c -R g+rwX /build; \
+RUN if [[ ! -f /tmp/v3.tgz ]] || [[ ${BOOTSTRAP} == "true" ]]; then \
+      # To only cache files from /latest/ folders, use --build-arg LATEST_ONLY=true 
+      if [[ ${LATEST_ONLY} == "true" ]]; then \
+        ./cache_artifacts.sh v3 --latest-only && chmod -R g+rwX /build; \
+      else \
```
amisevsk: This branch of the `if` does nothing, since `${LATEST_ONLY} == true` will cause non-latest metas to be deleted in an earlier step (line 76).

You'd need a separate arg to keep all metas and cache only latest.

nickboldt:
Agree, we don't need this extraneous logic in upstream... or downstream either.

PR #255, opened 2019-10-16
Release 7.3.0 version of che-plugin-registry 

ibuziuk:
### What does this PR do?
Release 7.3.0 version of che-plugin-registry  - https://github.com/eclipse/che/issues/14896 

- Bumping VERSION file
- Adding '7.3.0' versions for che-theia & machine-exec
- Setting 'latest' version for che-theia and machine-exec plugins to '7.3.0' (latest = latest release version)
- **Adding missing 'secure' & 'cookiesAuthEnabled' attributes to the latest version of the che-machine-exec plugin**
- Setting '7.3.0' version for container runtime images of the che remote plugins - `find . -name '*.yaml' -exec sed -i -e 's/:next/:7.3.0/g' {} \;`

PR #259, opened 2019-10-17
che #14896 Bumping version to '7.4.0-SNAPSHOT'. Adding '7.3.0' versions of che-theia and machine-exec plugins

ibuziuk:
NOTE: PR should be merged after resolving https://github.com/eclipse/che/issues/14918 and `che-theia` 7.3.0 release

### What does this PR do?
Version bump after 7.3.0 release - https://github.com/eclipse/che/issues/14896

- Bumping version to 7.4.0-SNAPSHOT
- Adding '7.3.0' versions of che-theia and machine-exec plugins
- Updating 'latest' verison of che-theia and machine-exec plugins to use '7.3.0' version of the container runtime image
- Fixing 'latest' version of 'che-machine-exec-plugin' (updating description, adding 'secure' and 'cookiesAuthEnabled' attributes, formatting)

PR #260, opened 2019-10-17
Update how latest versions of plugins are handled

amisevsk:
### What does this PR do?
Currently, the latest version of each plugin requires its own `meta.yaml` that is just a copy of the version we deem latest. As plugins are updated, the manual step in updating the latest version becomes error prone and can result in inconsistencies between two aliases for what should be the same plugin (see https://github.com/eclipse/che-plugin-registry/pull/255, where che-machine-exec had drifted out of sync between `latest` and `7.2.0`.

This PR modifies the way `latest` is handled; instead of a copy of the latest `meta.yaml`, the 'latest' directory for each plugin contains a file `latest.version`, which just contains the version number that should be used as "latest". During the docker build, a script reads these `latest.version` files and creates latest `meta.yaml`s automatically. 

To make the review a little simpler, there are two commits: the first replaces all current latest meta.yamls with the `latest.version` file, and the second adds the scripts to build an image with latest in place.


amisevsk:
@sr229 I've got one question about the code-server editor: the current `cdr/code-server/latest/meta.yaml` file references image `docker.io/chinodesuuu/coder:che-v2`, while `cdr/code-server/2.1523-vsc1.38.1/meta.yaml` references `docker.io/chinodesuuu/coder:2.1523-vsc1.38.1-che`. I see on dockerhub that these images are actually the same hash, so do we need to maintain the two versions of the plugin? Is your intention to have the `che-v2` represent a rolling release? 

If the two images will always be in sync, I'd prefer to eliminate one of the versions; if `che-v2` is meant to be a nightly/development build of the plugin, then I suppose latest should point to the fixed version `2.1523-vsc1.38.1/meta.yaml`, similarly to how the `next` version of `che-theia` is handled. As it stands I've left nothing changed, except breaking `che-v2` into its own version to accommodate the changes in this PR.

sr229:
> @sr229 I've got one question about the code-server editor: the current `cdr/code-server/latest/meta.yaml` file references image `docker.io/chinodesuuu/coder:che-v2`, while `cdr/code-server/2.1523-vsc1.38.1/meta.yaml` references `docker.io/chinodesuuu/coder:2.1523-vsc1.38.1-che`. I see on dockerhub that these images are actually the same hash, so do we need to maintain the two versions of the plugin? Is your intention to have the `che-v2` represent a rolling release?
> 
> If the two images will always be in sync, I'd prefer to eliminate one of the versions; if `che-v2` is meant to be a nightly/development build of the plugin, then I suppose latest should point to the fixed version `2.1523-vsc1.38.1/meta.yaml`, similarly to how the `next` version of `che-theia` is handled. As it stands I've left nothing changed, except breaking `che-v2` into its own version to accommodate the changes in this PR.

`che-v2` represents a "latest" stable rolling release but I'm fine on how you did it. The main reason I did `che-v2` was to reduce maintenance for the latest plugin version since pushing on a tag is easier and just have a versioned tag if they want to target a specific version

amisevsk:
@sr229 Makes sense, thank you.

build/scripts/generate_latest_metas.sh
```diff
@@ -0,0 +1,31 @@
...
+# Utility for automatically generating latest meta.yamls for plugins.
+#
+
+set -e
+
+for plugin_dir in v3/plugins/*/*; do
+  echo "Updating plugin ${plugin_dir}"
+  # Make sure every plugin has a 'latest.version' file
+  if [ ! -d "${plugin_dir}"/latest ] || [ ! -f "${plugin_dir}"/latest/latest.version ]; then
+    echo "    Missing latest.version: ${plugin_dir}/latest/latest.version"
```
nickboldt: I might prefer if we use `latest.txt` or `latest/version.txt` instead of `latest/latest.version` but :man_shrugging: doesn't change the functionality here.

build/scripts/generate_latest_metas.sh
```diff
@@ -0,0 +1,31 @@
...
+  fi
+  # Generate meta.yaml for latest version
+  version=$(cat "${plugin_dir}/latest/latest.version")
+  latest_meta="${plugin_dir}/${version}/meta.yaml"
+  if [ ! -f "${latest_meta}" ]; then
+    echo "    [ERROR]: version.latest specifies '$version' but no such meta.yaml is found"
+    echo "             expecting: '${plugin_dir}/${version}/meta.yaml'"
+  fi
+  echo "    Found latest meta ${plugin_dir}/${version}/meta.yaml"
+  yq -y '. + {version: "latest"}' "$latest_meta" > "${plugin_dir}/latest/meta.yaml"
```
nickboldt: I would also use a yq cleanup on `${plugin_dir}/${version}/meta.yaml` so both files are consistently formatted, including indents and quotes.

build/scripts/generate_latest_metas.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+#
+# Copyright (c) 2018-2019 Red Hat, Inc.
```
nickboldt: Wrong copyright date.

Also we should probably have a .gitignore for /latest/ folders so that we don't commit them to the repo, as they'll be generated from this point onward.

build/scripts/generate_latest_metas.sh
amisevsk: There are two issues with this that leave me inclined to not modify existing metas:
1. We can't just copy back over since we've changed the version, we'd have to change the version back. `yq` also doesn't support in-place editing (so we can't easily do `yq -i '.' $file` to get what we want).
2. Modifying other files in the repo like this would mean anyone submitting a PR would have to run the `generate_latest_metas.sh` script before opening a PR, and we don't have a way to validate that. If a user does not run the script locally outside of the build for their PR, then the next person who does run it will have modified files spuriously.

build/scripts/generate_latest_metas.sh
amisevsk: Good catch, will fix

build/scripts/generate_latest_metas.sh
nickboldt: Re: 1, you lost me... we're not changing the version, we're simply reformatting the yaml

Re: 2, yes. They should run it. If they forget, the next person will automatically fix their mistake. If neither person runs it, the yaml is still valid and we're no worse than today. 

But having the script do the copy from /latest/ back to /version/ would allow the OPTION for them to see that the two files differ and can then choose to commit the updated /version/meta.yaml so it matches with /latest/meta.yaml.

IMHO there's no downside here.

build/scripts/generate_latest_metas.sh
amisevsk: Done.

build/scripts/generate_latest_metas.sh
amisevsk: Done

amisevsk:
@nickboldt Fixed most of your comments:
- Added `v3/**/latest` to .gitignore
- Moved `latest.version` to `../latest.txt`
- Updated copyright headers

For rewriting existing meta.yamls, I feel it'll cause too many issues with spurious changes/PR process. Before including that functionality, I'd say we need a CI to make sure every PR has formatted their files using `yq`. Additionally, it might be impacted by autoformat from editors (e.g. vscode).

amisevsk:
Marked PR as [WIP] since it will conflict with adding 7.3.0 versions to registry: https://github.com/eclipse/che-plugin-registry/pull/259

amisevsk:
Rebased on master and updated `che-theia` and `che-machine-exec` latest to 7.3.0.

v3/plugins/cdr/code-server/latest.txt
```diff
@@ -0,0 +1 @@
+che-v2
```
sr229: I don't think its wise putting this on the rolling version since we want a stable plugin as much as possible.

amisevsk: Yeah I went to bed on Friday thinking that exact thing :)

I'll update this when I rebase, there's a few other plugins being added now so I want to do it only once.

@sr229 What do you think about naming the `che-v2` version `next`? That's kind of more in line with what we do with e.g. Theia currently, so it may be more understandable to a user. (Basically everything would be the same except the version that shows up in the Che dashboard).

sr229: Works for me 👍

amisevsk: Updated `che-v2` version to `next` and pointed `latest` at `2.1523-vsc1.38.1`

amisevsk:
@nickboldt rebased and fixed up new plugins

nickboldt:
:+1: merge with impunity

PR #263, opened 2019-10-22
Add redirect for resources to match one for plugins

amisevsk:
### What does this PR do?
Add redirect `/resources -> /v3/resources` to match the redirect for plugins.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -2,6 +2,7 @@ Header set Access-Control-Allow-Origin "*"
 Header set Access-Control-Allow-Headers "Authorization"
 DirectoryIndex README.md
 Redirect temp /plugins /v3/plugins
+Redirect temp /resources /v3/resources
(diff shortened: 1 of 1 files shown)


amisevsk:
> but who's using /resources/ ?? surely any reference to resources is from within the plugin reg, so this isn't needed?

The issue is that for the plugin broker, not including `/v3` in the path will work for normal registries and fail for offline registries.

Assuming the registry URL is `registry.com`, and Che is configured without the `/v3` the plugin broker will try to get e.g. `registry.com/plugins/redhat/java/latest.meta.yaml` and be redirected to the correct file (`registry.com/v3/plugins/redhat/java/latest.meta.yaml`). However, the devfile contains `relative:extension/resources/my-java.vsix`. Since the broker thinks the registry URL is `registry.com`, it attempts to get `registry.com/resources/my-java.vsix`, which fails without a redirect.

PR #265, opened 2019-10-23
Add list of external images to server when offline registry is built

amisevsk:
### What does this PR do?
Adds building a list of externally-referenced images to the offline build to make it easier for users to know which images need to be pulled into a private registry.

Required for https://github.com/eclipse/che/issues/14866


nickboldt:
Looks good, worked well on the crw plugin reg. Merged.

PR #273, opened 2019-10-25
Change the regex to replace registries that include colons

tomgeorge:
Signed-off-by: Tom George <tg82490@gmail.com>

### What does this PR do?
Fixes an issue where the plugin registry was not properly updating image references, causing the offline installation of che to try to pull limages from the wrong place

fixes https://github.com/eclipse/che/issues/14990


tomgeorge:
Closing in favor of https://github.com/eclipse/che-plugin-registry/pull/272

PR #276, opened 2019-10-28
Update VSCode-XML to 0.9.1

JPinkney:
### What does this PR do?

This PR updates the vscode-xml extension to 0.9.1.

Related issue: https://github.com/eclipse/che/issues/14992

Tested on minikube with deployed custom registry


nickboldt:
Does this updated version work with JDK 8 and if so, should we also update to this version in CRW plugin reg? If so, please submit a PR there: 

https://github.com/redhat-developer/codeready-workspaces/tree/master/dependencies/che-plugin-registry/v3/plugins/redhat/vscode-xml

nickboldt:
To be added to 7.3.1 tag and also merged to CRW 2 downstream

PR #264, opened 2019-10-23
#14927 remove node-debug (support for Node <...

nickboldt:
https://github.com/eclipse/che/issues/14927 remove node-debug (support for Node < 8) as Node 10 is LTS, Node 12 is Current, and Node 8 is old but still in maintenance. Do we REALLY need Node 4 & 6 support??

Signed-off-by: nickboldt <nboldt@redhat.com>

PR #279, opened 2019-10-29
Replace 'eclipse/che-theia-endpoint-runtime' image

evidolob:
### What does this PR do?
Replace `eclipse/che-theia-endpoint-runtime` image with `eclipse/che-remote-plugin-node` image

`eclipse/che-theia-endpoint-runtime` - is deprecated, see [che-theia PR](https://github.com/eclipse/che-theia/pull/521 )

### DIFF
--- a/v3/plugins/ms-vscode/node-debug/1.35.2/meta.yaml
+++ b/v3/plugins/ms-vscode/node-debug/1.35.2/meta.yaml
@@ -13,7 +13,7 @@ category: Debugger
 firstPublicationDate: '2019-06-20'
 spec:
   containers:
-  - image: "docker.io/eclipse/che-theia-endpoint-runtime:next"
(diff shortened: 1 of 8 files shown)


v3/plugins/ms-vscode/node-debug/1.35.2/meta.yaml
```diff
@@ -13,7 +13,7 @@ category: Debugger
 firstPublicationDate: '2019-06-20'
 spec:
   containers:
-  - image: "docker.io/eclipse/che-theia-endpoint-runtime:next"
+  - image: "docker.io/eclipse/che-remote-plugin-node:next"
```
nickboldt: we should just remove these entirely -- node-debug is for supporting Node versions < 8. See https://github.com/eclipse/che-plugin-registry/pull/264

PR #285, opened 2019-11-04
Remove /projects mount on codeserver, since mountSources is set to true

amisevsk:
### What does this PR do?
The codeserver editor has both `mountSources: true` and an explicit volume mount for `/projects`, which causes Che to create an invalid deployment: 

```
Deployment.apps "<workspace-id>.che-workspace-pod" is invalid: spec.template.spec.containers[0].volumeMounts[3].mountPath: Invalid value: "/projects": must be unique.
```

This PR removes the explicit `/projects` volume mount.

### DIFF
--- a/v3/plugins/cdr/code-server/2.1523-vsc1.38.1/meta.yaml
+++ b/v3/plugins/cdr/code-server/2.1523-vsc1.38.1/meta.yaml
@@ -26,10 +26,6 @@ spec:
          - exposedPort: 9000
      memoryLimit: "1024M"
      volumes:
-        - mountPath: '/projects'
-          name: 'projects'
-          ephemeral: true
-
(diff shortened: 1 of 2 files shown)


amisevsk:
cc: @sr229 Please double check -- I can't add you as an explicit reviewer.

benoitf:
I think that this PR https://github.com/eclipse/che-plugin-registry/pull/274 should also be updated

amisevsk:
Good catch @benoitf

sr229:
I think you can cherry pick my changes and just put it on this PR instead ;)

monaka:
@sr229 Cloud you review #305 ?

monaka:
I close this as the issue was fixed by #305.

PR #289, opened 2019-11-06
Add Jaeger as a plugin

objectiser:
### What does this PR do?

Recently wrote a [blog post](https://medium.com/jaegertracing/using-jaeger-with-eclipse-che-d0581c7da431) about using Jaeger to obtain distributed tracing information from services being developed. This was achieved by adding Jaeger to the [devfile](https://gist.github.com/objectiser/667615926a40d6cd8eb675859ddee1a1).

Thought it would be good if there was actually a plugin that could hide the details - hence this PR.

I have been trying to test it locally, but ran into [this issue](https://github.com/eclipse/che/issues/14517). So hoping someone else will be able to give it a quick test.


v3/plugins/redhat/jaeger/1.14/meta.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: v2
+publisher: redhat
+name: jaeger
+version: 1.14
+type: Che Plugin
+displayName: Jaeger
+title: Jaeger, Distributed Tracing System
+description: OpenTracing/OpenTelemetry compliant distributed tracing plugin for Eclipse Che
+icon: https://raw.githubusercontent.com/jaegertracing/documentation/master/static/img/jaeger-vector.svg
+category: Observability
```
objectiser: Wasn't sure whether to set this to `Other`, but thought might be good to have a new category for observability tooling.

v3/plugins/redhat/jaeger/1.14/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+      value: "9411"
+    ports:
+    - exposedPort: 16686
+    - exposedPort: 14250
+    - exposedPort: 14268
+    - exposedPort: 9411
+    - exposedPort: 5778
+    - exposedPort: 6831
+    - exposedPort: 6832
+    memoryLimit: "128Mi"
```
objectiser: Wasn't sure whether to add this, or leave it for the user to specify in the devfile when using the plugin?

v3/plugins/redhat/jaeger/1.14/meta.yaml
objectiser: Have changed to `Other` due to build failure - but might be worth considering a new category :)

l0rd:
Thanks @objectiser for the PR. This is not really a plugin in the sense of a vscode extension or theia plugin. It's more of a workspace runtime rather then a plugin. What I mean is that if we add jaeger as a plugin we could add postgres as well. 

But on the other end there is no doubt that it extends a che workspace capabilities. 

That said I am not sure if it should live in the plugin registry or not.  @tsmaeder @benoitf @skabashnyuk @davidfestal @evidolob @slemeur what's your point of view on this?

objectiser:
@l0rd Understand your point - the reason I thought about creating it was because the `dockerimage` approach was defined in the [components section](https://gist.github.com/objectiser/667615926a40d6cd8eb675859ddee1a1) along side,
```
  -
    type: chePlugin
    id: redhat/java/latest
    memoryLimit: 1400Mi
```
which is defined as a plugin - and is also a workspace runtime (I assume - haven't looked into the details of the plugin).

In general it provides a more concise representation in the devfile - so even if plugin is not the right term, it would be good if there was an equivalent capability to enable users to easily compose a workspace.

PR #291, opened 2019-11-08
Add Quarkus plugin

svor:
Signed-off-by: svor <vsvydenk@redhat.com>

### What does this PR do?
Registers Quarkus plugin. The plugin will be run in a container with  `docker.io/eclipse/che-remote-plugin-runner-java8:next` image and include next VS Code extensions:
- Java JDT.LS
- vscode-java-debug
- vscode-quarkus


v3/plugins/redhat/quarkus/1.1.1/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+  containers:
+    - image: "docker.io/eclipse/che-remote-plugin-runner-java8:next"
+      name: vscode-quarkus
+      memoryLimit: "1500Mi"
+      volumes:
+      - mountPath: "/home/theia/.m2"
+        name: m2
+  extensions:
+    - https://github.com/microsoft/vscode-java-debug/releases/download/0.23.0/vscode-java-debug-0.23.0.vsix
+    - https://download.jboss.org/jbosstools/static/jdt.ls/stable/java-0.50.0-1825.vsix
```
tsmaeder: I don't think it makes sense to have a separate Quarkus extensions. We need to include this support in the Java extension. Otherwise, we'll have to have a separate plugin for all combinations of technologies. We should be mix-and-match

v3/plugins/redhat/quarkus/1.2.0/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+name: quarkus
+version: 1.2.0
+type: VS Code extension
+displayName: Quarkus Tools
+title: Quarkus Tools by Red Hat
+description: This plug-in provides support for Quarkus development via a Quarkus language server, Java jdt.ls and Java Debugger.
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/redhat-developer/vscode-quarkus
+category: Language
+firstPublicationDate: "2019-11-25"
```
nickboldt: This date is now wrong :)

v3/plugins/redhat/quarkus/1.2.0/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+spec:
+  containers:
+    - image: "docker.io/eclipse/che-remote-plugin-runner-java8:next"
+      name: vscode-quarkus
+      memoryLimit: "1500Mi"
```
nickboldt: Per discussion in https://github.com/eclipse/che-devfile-registry/pull/144#event-2900270521 is 1.5G enough memory?

v3/plugins/redhat/quarkus/1.2.0/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+      volumes:
+      - mountPath: "/home/theia/.m2"
+        name: m2
+  extensions:
+    - https://github.com/microsoft/vscode-java-debug/releases/download/0.23.0/vscode-java-debug-0.23.0.vsix
```
nickboldt: Should we also update https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/java11/0.50.0/meta.yaml#L22 and https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/java8/0.50.0/meta.yaml#L22 and https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/java/0.50.0/meta.yaml#L22 to use this newer version of the debug plugin?

v3/plugins/redhat/quarkus/1.2.0/meta.yaml
svor: updated it a bit :)

v3/plugins/redhat/quarkus/1.2.0/meta.yaml
svor: i didn't try big projects, but for https://github.com/quarkusio/quarkus-quickstarts.git samples 1.5G is enough

v3/plugins/redhat/quarkus/1.2.0/meta.yaml
svor: I've created an issue for that. https://github.com/eclipse/che/issues/15560

v3/plugins/redhat/quarkus/1.2.0/meta.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+publisher: redhat
+name: quarkus
```
sunix: I would name the plugin `quarkus-java8` because it is bringing quarkus ext + java and java8 runtime. Would make sence to avoid confusion like `do I need to install java8 or java11 plugin if I already have quarkus ?

nickboldt: I've been assuming that for CRW 2.1 we'd need a JDK11 runtime for use with quarkus.

Are you saying that quarkus works with BOTH JDK8 and 11, and therefore we will need two sidecar containers, one per JDK flavour?

svor: I've updated `name` and `displayName`

PR #290, opened 2019-11-07
Use nightly version che-machine-exec from centos Ci.

AndrienkoAleksandr:
### What does this PR do?
Use nightly version che-machine-exec from centos Ci. Remove next tag from Codenvy CI.

### Referenced issue:
https://github.com/eclipse/che/issues/14887

Signed-off-by: Oleksandr Andriienko <oandriie@redhat.com>

PR #295, opened 2019-11-13
Use nightly version che-machine-exec from centos Ci. (#290)

AndrienkoAleksandr:
### What does this PR do?
* Use nightly version che-machine-exec from centos Ci.
Use nightly version che-machine-exec from centos Ci.

### Referenced issue
https://github.com/eclipse/che/issues/14887

### DIFF
--- a/v3/plugins/eclipse/che-machine-exec-plugin/next/meta.yaml
+++ b/v3/plugins/eclipse/che-machine-exec-plugin/next/meta.yaml
@@ -10,6 +10,9 @@ description: Che Plug-in with che-machine-exec service to provide creation termi
 icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
 repository: https://github.com/eclipse/che-machine-exec/
 firstPublicationDate: "2019-02-05"
+deprecate:
+  automigrate: true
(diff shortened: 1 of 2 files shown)


ibuziuk:
@AndrienkoAleksandr I'm worrying that this might have side-effects if the `next` version will not exist anymore in 7.3.x

AndrienkoAleksandr:
> @AndrienkoAleksandr I'm worrying that this might have side-effects if the next version will not exist anymore in 7.3.x

Done I saved next tag.

ibuziuk:
@AndrienkoAleksandr maybe we should point to the `nighly` image from the `next` version of the plugin?

AndrienkoAleksandr:
> @AndrienkoAleksandr maybe we should point to the nighly image from the next version of the plugin?

If it is ok for you, I can do it in this pr.

AndrienkoAleksandr:
>  @ibuziuk maybe we should point to the nighly image from the next version of the plugin?
> @AndrienkoAleksandr  Also I can mark next like deprecated.

Done.

ibuziuk:
@amisevsk could you please take a look, I believe we should not have side effect for airgap

PR #297, opened 2019-11-14
Release Che 7.3.2 version of the che-plugin-registry

ibuziuk:
### What does this PR do?

- Bumping version to 7.3.2
- Adding 7.3.2 versions of 'che-theia' and 'che-machine-exec' and updating `latest.txt` of those plugins
- Setting '7.3.1' version for container runtime images of the che remote plugins via `find . -name '*.yaml' -exec sed -i -e 's/:7.3.1/:7.3.2/g' {} \;`

issue https://github.com/eclipse/che/issues/15107


v3/plugins/eclipse/che-machine-exec-plugin/7.3.0/meta.yaml
```diff
@@ -24,6 +24,6 @@ spec:
         cookiesAuthEnabled: true
   containers:
    - name: che-machine-exec
-     image: "quay.io/eclipse/che-machine-exec:7.3.1"
+     image: "quay.io/eclipse/che-machine-exec:7.3.2"
```
amisevsk: Not sure on the process here, but should `che-machine-exec/7.3.0` use the `7.3.2` image, or is this a side-effect of the sed replacement? I'm not sure how versioning works with machine-exec/che-theia, since e.g. `7.2.0` doesn't get its image updated, but e.g. `typescript` has all versions updated. This update makes e.g. che-machine-exec `7.3.0`, `7.3.1`, and `7.3.2` identical.

I've got a similar question for `che-machine-exec/7.3.1` and `che-theia/7.3.0`, `che-theia/7.3.1`. At first glance it looks like the images used should match the bugfix release of Che in their version.

ibuziuk: oh, good catch, this should not be changed I suppose

ibuziuk: fixed

ibuziuk:
@amisevsk fixed. Does it look good now ?

PR #305, opened 2019-11-29
Fix issues on codeserver editor plugins.

monaka:
### What does this PR do?

Fixes #274 #285 .
I cherry-picked from them and added one more fix.


monaka:
@amisevsk Could you review this ?

PR #306, opened 2019-12-02
fix(theia-endpoint): Use tagged images for the theia plug-ins sidecar

benoitf:
### What does this PR do?
As we don't need anymore to follow Che-Theia lifecycle as the endpoint binary is not anymore inside the container, we can use fixed tagged images for plug-ins sidecars.

Part of https://github.com/eclipse/che/issues/15109

Change-Id: I1f26ffe38c8cb75d8048e99d11b3e5b12790cc53

### DIFF
--- a/v3/plugins/che-incubator/theia-dev/0.0.1/meta.yaml
+++ b/v3/plugins/che-incubator/theia-dev/0.0.1/meta.yaml
@@ -19,7 +19,7 @@ spec:
       protocol: http
   containers:
    - name: theia-dev
-     image: "docker.io/eclipse/che-theia-dev:nightly"
+     image: "docker.io/eclipse/che-theia-dev:next"
      commands:
        - name: uname
          workingDir: "$(project)"
--- a/v3/plugins/broadcommfd/cobol-language-support/0.8.3/meta.yaml
+++ b/v3/plugins/broadcommfd/cobol-language-support/0.8.3/meta.yaml
@@ -11,7 +11,7 @@ repository: https://github.com/eclipse/che-che4z-lsp-for-cobol
 category: Language
 spec:
   containers:
-    - image: eclipse/che-remote-plugin-runner-java8:7.3.0
+    - image: "quay.io/eclipse/che-sidecar-java:8-fac2415"
       memoryLimit: 512Mi
   extensions:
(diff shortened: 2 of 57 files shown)


ibuziuk:
@benoitf could you please clarify how this would affect the plugin-registry release cyccle? am I correct that the only plugins that would need to be updated during the release process are machine-exec and che-theia ?

benoitf:
@ibuziuk correct, it will simplify a lot the registry release.
for che-theia project, only che-theia needs to be added with correct version

and yes, we're still releasing as well machine-exec at every release but it may be optional as well (if we don't follow same lifecycle than eclipse che) but it's out of scope.

v3/plugins/che-incubator/theia-dev/0.0.1/meta.yaml
```diff
@@ -19,7 +19,7 @@ spec:
       protocol: http
   containers:
    - name: theia-dev
-     image: "docker.io/eclipse/che-theia-dev:nightly"
+     image: "docker.io/eclipse/che-theia-dev:next"
```
ibuziuk: is this nightly / next switch intentional? what is the reason for changing it?

benoitf: @ibuziuk yes it's intentional, I've used a separate commit for it
https://github.com/eclipse/che-plugin-registry/pull/306/commits/b4860a53846b2830f83754cd819fe111339847e4

nightly was not built since 9months: https://hub.docker.com/r/eclipse/che-theia-dev/tags?page=1&name=nightly

as it was switched to next

benoitf:
@amisevsk I've myself the same question. 8 is an alias to the latest image. But in the past for example, the images were quite changed internally (like to make separate endpoint process, etc) so I'm unsure that if che 7.0.0 was still using the `:8` tag it will work.

Another idea that I had was to use alias like `che-sidecar-java:8` during the development process/master branch and only when we do a release of che-plugin-registry, we set the fixed version (with sha1 suffix).
Then it's more a question for @ibuziuk , if the release could do that, if it's a good or bad idea.

amisevsk:
@benoitf That makes sense, I mostly wanted clarification. Ideally the release process would do something like what you describe (or maybe even just set a SHA hash instead of tag), but I think it's too manual and complicated as it stands so I'm :+1: on merging as-is.

PR #309, opened 2019-12-03
#14573 do not reference MS VSCode...

nickboldt:
#14573 do not reference MS VSCode marketplace directly; instead use v2.7.7 which is in GH


nickboldt:
See issue #15327 -- should we just REMOVE 2.7.6 in favour of 2.7.7, or add 2.7.7 and leave 2.7.6 hanging around even though it's in violation of #14573 ?

themr0c:
Then keep version 2.7.6 for some time and add version 2.7.7 now as the new default?

nickboldt:
Updated to keep 2.7.6 (but deprecate it). 

Note that the docs for the deprecation give non-example of `migrateTo: another_publisher/another_plugin_id:1.0.0`  but an actual working example shows we use `migrateTo: redhat/vscode-apache-camel/latest`, not using `:tag` syntax

PR #319, opened 2019-12-10
Add 7.3.2 and 7.3.3 versions of che-theia/che-machine-exec

amisevsk:
### What does this PR do?
Add the 7.3.2 and 7.3.3 bugfix version of che-machine-exec and che-theia to master branch to ensure compatibility when updating Che.

Metas are copied from 7.3.3 tag.

Fixes https://github.com/eclipse/che/issues/15444 for future releases.

### DIFF
--- a/v3/plugins/eclipse/che-machine-exec-plugin/7.3.2/meta.yaml
+++ b/v3/plugins/eclipse/che-machine-exec-plugin/7.3.2/meta.yaml
@@ -1,0 +1,29 @@
+apiVersion: v2
+publisher: eclipse
+name: che-machine-exec-plugin
+version: 7.3.2
+type: Che Plugin
+displayName: Che machine-exec Service
+title: Che machine-exec Service Plugin
+description: Che Plug-in with che-machine-exec service to provide creation terminal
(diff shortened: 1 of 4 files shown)


nickboldt:
Alternative idea...

can the plugin broker, if it is asked for a version that doesn't exist... simply grab /latest as a fallback behaviour? Better that than:
* forced hardcoding of old versions / dead boilerplate code
* breaking compatibility when we forget to push 7.y.z+1 versions into master

ibuziuk:
> Should these old versions be marked deprecated/auto-update to the latest? Compatibility is good but so is staying current w/ 7.x.next

I believe we should follow the flow we had since the very beginning related to backward compatibility.

> can the plugin broker, if it is asked for a version that doesn't exist... simply grab /latest as a fallback behaviour? Better that than:  

yes, there is an rh-che issue for that https://github.com/redhat-developer/rh-che/issues/1396 
I thik we need to create an upstream issue for that. In any case I think this should not block this PR from merging

nickboldt:
Agreed, merged.

PR #321, opened 2019-12-11
fix(camel-k): Update images to the latest one

benoitf:
### What does this PR do?
entrypoint was not executable
Fixes https://github.com/eclipse/che/issues/15455

Change-Id: Id00a2a710b626cd03158d24e66bac31b7c02e9bd

### DIFF
--- a/v3/plugins/redhat/vscode-camelk/0.0.10/meta.yaml
+++ b/v3/plugins/redhat/vscode-camelk/0.0.10/meta.yaml
@@ -12,7 +12,7 @@ category: Language
 firstPublicationDate: '2019-07-10'
 spec:
   containers:
-  - image: "quay.io/eclipse/che-sidecar-camelk:0.0.10-966607d"
+  - image: "quay.io/eclipse/che-sidecar-camelk:0.0.10-70f9d17"
     name: vscode-camelk
     memoryLimit: "1G"
   extensions:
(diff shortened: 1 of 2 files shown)


apupier:
to help reviewing the PR, here is the link to the tags of the images updated https://quay.io/repository/eclipse/che-sidecar-camelk?tag=latest&tab=tags

nickboldt:
My opinion is that the 0.0.9 one should be marked deprecated with auto update to the 0.0.10 one so you don't have to bother rebuilding images for old plugins.

But otherwise I have no opinion here. +0.1, looks OK.

benoitf:
@nickboldt yes but it was failing to start container for 0.0.9

benoitf:
@apupier is it ok to deprecate 0.0.9 in a new PR ?

nickboldt:
but wouldn't the auto-update cause the plugin broker to pull the 0.0.10 plugin instead of the 0.0.9 one, then the 0.0.10 container? cc: @amisevsk (I'm trying to understand how the deprecate/auto-update actually works in terms of pulled containers after reading meta.yaml

benoitf:
ok can do deprecate stuff in a new PR

amisevsk:
@nickboldt I don't think the broker will do anything with those fields as it currently stands; the update has to come from the server-side. From what I've seen (I'm very ready to be wrong on this one), the deprecate support is only a warning on the workspace config tab.

nickboldt:
@amisevsk then what are these for?
```
  autoMigrate:         # boolean
  migrateTo:           # new org/plugin-id/version, e.g. redhat/vscode-apache-camel/latest
```

If it's just a warning then how is the `migrateTo` useful?

amisevsk:
@nickboldt -- as I mentioned, I'm happy to be wrong. On the other hand, I'm able to create and update workspace with the currently deprecated `camel-tooling/vscode-apache-camel` plugin. I've opened issue https://github.com/eclipse/che/issues/15461

PR #322, opened 2019-12-12
Che4z-1.1.0

filipkroupa:
cobol-language-support version 0.9.1
introduced debugger-for-mainframe and hlasm-language-support

### What does this PR do?

Che4z plugins: add updated version of COBOL Language Support, introduce Debugger for Mainframe and HLASM Language Support

Targetting [Milestone 7.6.0](https://github.com/eclipse/che/milestone/107)


filipkroupa:
@amisevsk I agree with you. There is an effort to prepare demo projects for the language plugins based on internal COBOL and HLASM sources. Other plugins would require servers to connect to, which we do not have exposed at the moment.

PR #325, opened 2019-12-16
feat(core): upgrade Language Support for Camel to 0.0.20

apupier:

### What does this PR do?

it is bringing a newer version of Language Support which is including several noticeable new features: https://developers.redhat.com/blog/2019/12/16/vs-code-language-support-for-apache-camel-0-0-20-release/


amisevsk:
Some brief testing of the new version in the camel-springboot devfile suggests that a revised limit of `260Mi` will work, but I didn't stress the camel lang server very much. This should also fit within the default 3Gi quota on che.openshift.io.

apupier:
updated to 200Mi

amisevsk:
@apupier Sorry, I wasn't clear. The `512MI` limit in this repo is generally okay (maybe could be reduced to 300Mi to make it easier to add the plugin). 

With 200Mi, though, I got an OOMKIlled error when starting the apache-camel-springboot stack on nightly Che, so I think 200Mi is not enough and had to bump it to 260Mi to get it to start without issue. This change needs to happen in the *devfile registry* instead of this one.

My suggestion would be:
1. Che Plugin Registry: Update this PR to set limit at `300Mi` or back to `512Mi`; I think 200Mi is too low as a default.
2. Che Devfile Registry ([this devfile](https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/apache-camel-springboot/devfile.yaml#L19)): Set limit to 260Mi for the redhat/vscode-apache-camel/latest plugin

apupier:
updated this PR to use 300Mi by default and changed in the devfile registry: https://github.com/eclipse/che-devfile-registry/pull/173

PR #338, opened 2020-01-08
Release 7.7.0 version of the che-plugin-registry

ibuziuk:
### What does this PR do?

Related issue - https://github.com/eclipse/che/issues/15609

- Bumping VERSION file to `7.7.0`
- Adding '7.7.0' versions of the che-theai and machine-exe and bumping the latest.txt files of those plugins
- Bumping version of the container runtime image of the `theia-dev` plugin
- Fixing the container runtime image for the quarkus plugin


ibuziuk:
@benoitf could you please take a look at that one. AFAIK, currently the only version of container runtime image that should be bumped from `next` to `7.7.0` is theia-dev

VERSION
```diff
@@ -1 +1 @@
-7.7.0-SNAPSHOT
+7.7.0
```
amisevsk: Should this change be merged to master or updated in the release branch?

ibuziuk: yeah, good catch - changing base to 7.7.x

PR #340, opened 2020-01-08
use vscode-kubernetes-tools-1.0.4 instead of...

nickboldt:
use vscode-kubernetes-tools-1.0.4 instead of 1.0.0 for consistency with latest
not sure if this change should precipitate creation of a 0.0.10a plugin version

Change-Id: I699c7f23bd194479403719a5d7d6d4fc4bd23cda

### DIFF
--- a/v3/plugins/redhat/vscode-camelk/0.0.10/meta.yaml
+++ b/v3/plugins/redhat/vscode-camelk/0.0.10/meta.yaml
@@ -17,4 +17,4 @@ spec:
     memoryLimit: "1G"
   extensions:
   - https://download.jboss.org/jbosstools/vscode/stable/vscode-camelk/vscode-camelk-0.0.10-153.vsix
-  - https://github.com/Azure/vscode-kubernetes-tools/releases/download/1.0.0/vscode-kubernetes-tools-1.0.0.vsix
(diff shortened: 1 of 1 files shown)


ibuziuk:
@nickboldt looks like you simply ignore the comments :/

ibuziuk:
@nickboldt oh, sorry my bad I meant *after* :man_facepalming:

nickboldt:
oops sorry. I followed your instructions :(

ibuziuk:
@nickboldt totally my bad, but anyway this does not look critical this will be part of 7.8.0

apupier:
I think the eclipse/che-sidecar-camelk needs to be rebuilt and updated because there is a new one for kubernetes. it has been upgraded https://github.com/eclipse/che-plugin-registry/blob/96f96bb417ee42e3c04bb5657f0272e7222e614e/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/1.0.0/meta.yaml#L18 https://github.com/eclipse/che-plugin-registry/blob/96f96bb417ee42e3c04bb5657f0272e7222e614e/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/1.0.4/meta.yaml#L15

apupier:
(even though it doesn't seems to cause functional trouble)

PR #345, opened 2020-01-13
Upgrade Camel Language support to 0.0.22 eclipse/che#15675

apupier:

### What does this PR do?

TODO


amisevsk:
> Should previous version(s) be marked deprecated so existing users will seamlessly be updated to latest ?

If users are using `latest`, they will be seamlessly upgraded. If they chose a specific version I see no reason to deprecate it here. In general, setting deprecated every time a plugin is updated seems like a bad policy to me (why even keep old versions at that point?).

nickboldt:
> why even keep old versions at that point?

Yes, Why don't we deprecate things and then remove them 3 sprints later, as per the deprecation/removal policy?

I mean these are 0.0.x versions. Not even 0.x from a stability/GA quality perspective. Incubating++ versions shouldn't need to be kept forever.

apupier:
I'm balanced between deprecating or not.
Given that these are 0.0.x versions that we are providing and that I think there is no breaking changes, I updated the PR to deprecate the old versions.
Having a clearer policy at Che (and CodeReady Workspaces) might help to choose in the future.

PR #346, opened 2020-01-13
Upgrade wsdl2rest extension to 0.0.11 eclipse/che#15674

apupier:

### What does this PR do?

TODO


apupier:
I'm balanced between deprecating or not.
Given that these are 0.0.x versions that we are providing and that I think there is no breaking changes, I updated the PR to deprecate the old versions.
Having a clearer policy at Che (and CodeReady Workspaces) might help to choose in the future.

amisevsk:
I agree that more policy on deprecate + automigrate is needed, if our policy is to always deprecate + automigrate on update, we may as well only support `latest` as a version.

PR #356, opened 2020-01-24
Add Metals, the Scala language server

sideeffffect:
Add support for Scala via the Metals language server, based on the already existing VS Code extension

https://marketplace.visualstudio.com/items?itemName=scalameta.metals

/cc @ibuziuk 

Signed-off-by: sideeffffect ondra.pelech@gmail.com


v3/plugins/scalameta/metals/1.7.2/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  containers:
+    - image: "quay.io/eclipse/che-sidecar-java:8-0cfbacb"
+      name: vscode-metals
+      memoryLimit: "1500Mi"
+      volumes:
+      - mountPath: "/home/theia/.m2"
+        name: m2
+  extensions:
+    - https://marketplace.visualstudio.com/_apis/public/gallery/publishers/scalameta/vsextensions/metals/1.7.2/vspackage/
```
amisevsk: Microsoft forbids downloading extensions from the vscode marketplace when they're not used in VS Code (see https://github.com/eclipse/che/issues/14573). From looking at the github repo, it seems they don't publish vsix releases, so I don't know the best option here.

sideeffffect: temporary workaround provided -- I'll talk to the upstream developers about it

nickboldt: Alternatively, if you can't get them to publish to GitHub or some other location, you could ask @mkuznyetsov  about how to build the vsix + source tarball and publish to download.jboss.org where we have all the 3rd party and RH-built extensions used in Che and CRW.

v3/plugins/scalameta/metals/1.7.2/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  containers:
+    - image: "quay.io/eclipse/che-sidecar-java:8-0cfbacb"
+      name: vscode-metals
+      memoryLimit: "1500Mi"
+      volumes:
+      - mountPath: "/home/theia/.m2"
+        name: m2
+  extensions:
+    - https://github.com/sideeffffect/metals-vscode/releases/download/1.7.2/scalameta.metals-1.7.2.vsix
```
amisevsk: We generally want to avoid pulling vsixes from forked repos. I'd prefer a more stable vsix source.

sideeffffect: I will talk to the extension developers after we test that it's capable of running in Che (https://github.com/eclipse/che/issues/15820)

amisevsk: Thanks @sideeffffect for your contributions!

v3/plugins/scalameta/metals/1.7.2/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+displayName: Scala (Metals)
+title: Scala language server with rich IDE features
+description: This plug-in provides support for Scala development via the Metals language server.
+icon: https://rawcdn.githack.com/OlegIlyenko/scala-icon/c453c6f67dfcc1dbf0ca1e437d32d04f1d710983/scala-icon.svg
+repository: https://github.com/scalameta/metals-vscode
+category: Language
+firstPublicationDate: "2020-01-24"
+spec:
+  containers:
+    - image: "quay.io/eclipse/che-sidecar-java:11"
```
amisevsk: Our standard practice for other plugins is to use the specific build tags in quay.io, i.e. `11-f76ca45` instead of `11`. @benoitf can you confirm?

v3/plugins/scalameta/metals/1.7.2/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+      name: vscode-metals
+      memoryLimit: "1500Mi"
+      volumes:
+      - name: sbt
+        mountPath: /home/theia/.sbt
+      - name: ivy2
+        mountPath: /home/theia/.ivy2
+      - name: coursier
+        mountPath: /home/theia/.cache/coursier
```
amisevsk: This volume mount is causing issues for us due to how subpaths are created in OpenShift. Since `/home/theia/.cache` does not exist, it is created when mounting the `coursier` folder. However, it is created with `755` permissions with root owner and group, and so is not writable by any process in the container.

v3/plugins/scalameta/metals/1.7.2/meta.yaml
sideeffffect: done

v3/plugins/scalameta/metals/1.7.2/meta.yaml
sideeffffect: I shortened `/home/theia/.cache/coursier` to `/home/theia/.cache` and things seem to be working

sideeffffect:
I have confirmed that the extension works, when the root of a Scala sbt project is opened as workspace :tada: 
image

sideeffffect:
can we merge this, so that it's available for testing for other people (most notably the Metals and VS Code extension developers)?

v3/plugins/scalameta/metals/1.7.2/meta.yaml
amisevsk: Yeah, that should fix the startup issue. The problem, however, is that now we're persisting the entire user `.cache` directory, which could potentially cause other problems. I think a better solution is to add volumes for both `.cache/metals` and `.cache/coursier` if those caches need to be persisted.

What are these caches used for? If we don't have them persistent, how negatively is performance impacted?

v3/plugins/scalameta/metals/1.7.2/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - image: "quay.io/eclipse/che-sidecar-java:11-f76ca45"
+      name: vscode-metals
+      memoryLimit: "1500Mi"
+      volumes:
+      - name: sbt
+        mountPath: /home/theia/.sbt
+      - name: ivy2
+        mountPath: /home/theia/.ivy2
+      - name: cache
+        mountPath: /home/theia/.cache
```
amisevsk: See my comment elsewhere, I think a better approach is to have volumes for `.cache/metals` and `.cache/coursier` if they need to be persisted, since linking the `.cache` directory in general to a PVC could cause problems.

v3/plugins/scalameta/metals/1.7.2/meta.yaml
sideeffffect: `.cache/coursier` not being persisted means that you would have to download all packages again, every time

v3/plugins/scalameta/metals/1.7.2/meta.yaml
sideeffffect: so what am I supposed to do, since `mountPath: /home/theia/.cache/coursier` didn't work?

amisevsk: Sorry, I'm not being very clear. Mounting `mountPath: /home/theia/.cache/coursier` worked for coursier, but failed because we couldn't create `/home/theia/.cache/metals`. It should work without problem if we have 
```yaml
      - name: coursier
        mountPath: /home/theia/.cache/coursier 
      - name: metals
        mountPath: /home/theia/.cache/metals 
```

sideeffffect: done

sideeffffect:
I've added the additional `mountPath`s and things seem to be working fine
but does that mean, that whenever any process now tries to create any other new directory/file in `~/.cache`, that it won't work/will crash?
that seems rather unpractical -- this could render this new Scala Che plugin practically unusable
:confused:

amisevsk:
@sideeffffect You make a good point -- I think the real solution would be to update the sidecar image. I can open a PR for that. 

What's the downside for not persisting the `coursier` and `metals` directories?

sideeffffect:
Not caching `~/.cache/{coursier, metals, bloop}` means that every time, all the packages, Metals (Scala language server) and Bloop (compilation server) have to be downloaded on every Che start, again, and again and again :frowning_face: 

> the real solution would be to update the sidecar image. I can open a PR for that

I probably still don't really understand the issue at hand, but if you thing that would resolve the issue, please do that :+1: 
Btw, what would be the nature of that change? Would you mind CCing me on that PR?

amisevsk:
@sideeffffect Do you know if it's possible to specify a different location for the `{coursier, metals, bloop}` folders? Additionally, is there a way to standalone install the required tools into an image before launching the language server (I found docs on coursier and bloop but metals is pretty scant).

sideeffffect:
> different location

> standalone install

it would be for the best, if we wouldn't have to worry about that, and just delegate to Metals to take care of that. Is that really necessary? Why not just mount whole `~/.cache`? We've already verified that works.

Or fix the issue with mounting volumes/creating subdirectories?

IMHO until the subdir issue is fixed, `~/.cache` is probably the mos suitable solution.

amisevsk:
@sideeffffect The reason I'm reluctant to add `.cache` as a PVC is that it could be used by other programs as well -- it's not specific to metals, coursier, etc. Something like yarn will fill this PVC very quickly, and we don't want to have to think about all possible workspace configurations.

sideeffffect:
would it help, if the `~/.cache` directory already existed in the base image?

sideeffffect:
anyway, we can tune these things after more people test this
I've changed it back to `~/.cache/{coursier, metals, bloop}` and tested that it indeed works fine
I think we can merge this PR

sideeffffect:
ping :smiley_cat:

amisevsk:
Hey @sideeffffect, sorry I left this unattended. In general the PR is good, but we still want to avoid merging a plugin that depends on releases from a fork instead of an "official" source. This could cause various issues around updating the plugin, etc. I'm looking into publishing built vsixes like we do for some other plugins that have this problem. 

Additionally, I'd like to build a custom sidecar container for scala to avoid having to download metals on startup (or every start, when we're running in ephemeral mode), but need to find some time to do that.

sideeffffect:
> we still want to avoid merging a plugin that depends on releases from a fork instead of an "official" source

I've removed those extensions, turns out, they're not necessary :tada: 
Do you think we're now ready to merge, to have at least a minimal viable version in master?

sideeffffect:
interesting,
> * The language server seems to fail to initialize on the first run (it has to download metals, and seems to log a series of errors afterwards)

I've never witnessed such behavior

> On second start, the language server works as expected, except you have to first open the project's folder in Theia (i.e. you have to open `/projects/console-java-simple` instead of `/projects`, as the lang server will complain about no `build.sbt`.

I always do this (open the `/projects/console-java-simple` as a workspace, since that's where `build.sbt` is) as the first thing, and everything works fine, from the very start.

amisevsk:
@sideeffffect The issue is that having to open the subproject at workspace start is a significant departure from the normal flow -- we don't expect users to know to do that, so they would just see a non-working plugin. All of the other language servers (e.g. Java) will correctly detect the pom.xml in the subfolder.

I don't know enough about how this is managed on the Theia side to be able to help there, though.

sideeffffect:
let's ask the authors :)
@olafurpg @tgodzik @gabro is it normal that Metals doesn't work (or work well) unless specifically the directory with `build.sbt` is opened -- in contrast to a situation when `build.sbt` along with the rest of the Scala project is in a subdirectory of the currently opened Workspace? We're trying to add Metals VS Code plugin to Eclipse Che, but we're not sure if it's working correctly.
Thanks for any help :pray: 

@amisevsk suggests that other language servers can work in that situation:
> All of the other language servers (e.g. Java) will correctly detect the pom.xml in the subfolder.

tgodzik:
> let's ask the authors :)
> @olafurpg @tgodzik @gabro is it normal that Metals doesn't work (or work well) unless specifically the directory with `build.sbt` is opened -- in contrast to a situation when `build.sbt` along with the rest of the Scala project is in a subdirectory of the currently opened Workspace? We're trying to add Metals VS Code plugin to Eclipse Che, but we're not sure if it's working correctly.
> Thanks for any help 
> 

We need `build.sbt` at the root of the workspace to detect if it's an sbt workspace. I can confirm it's the current behaviour.

> @amisevsk suggests that other language servers can work in that situation:
> 
> > All of the other language servers (e.g. Java) will correctly detect the pom.xml in the subfolder.

I don't think it's a correct behavior really, but might be correct for Maven. The heuristic here can have issues with a lot of corner cases. It's not obvious on how deep we should look or what to do with multiple separate projects. Some sbt builds might be quite complex.

As an alternative we could provide an additional setting for workspace root, but in most editors it's just easiest to open the right directory than to tinker with the settings. We can discuss it on the Metals issue tracker if needed, so that everyone can pitch in.

amisevsk:
Thank you very much @tgodzik for the input, I'll have to look into how other language plugins handle this, I admit I'm not familiar with the details there.

amisevsk:
@tgodzik After a brief dive into this, here's what I've found:

In the JDT.ls language server, we [specifically handle](https://github.com/eclipse/eclipse.jdt.ls/blob/6c6db8033d7ed02148a15e5eb886d776774a99da/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/InitHandler.java#L95-L117) multi-folder workspaces using the `workspaceFolders` parameter in the initialization parameters ([LSP doc link](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_workspaceFolders))

On the metals side, it appears that the pre LSP-3.6.0 spec is still assumed: https://github.com/scalameta/metals/blob/80868dad3d0b30aace2d460be093de9a4c9ad36b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L154

Does this make sense? I've opened https://github.com/scalameta/metals-feature-requests/issues/87 for further discussion on the Metals side.

tgodzik:
@amisevsk och yeah, we plan to fix it, but haven't had the time, since this requires a bit of refactor. Can't guarantee, this will be tackled any time soon. :disappointed:

olafurpg:
Multiple workspace roots is a desirable feature but I don't think it should be blocking this PR. There's nothing inherent blocking Metals from supporting multiple workspace roots, just somebody with time and motivation to work on it :)

amisevsk:
@olafurpg in terms of this PR, Che runs with multiple roots by default, so having to change the Theia folder is a very unintuitive flow. I would personally consider this a blocker from a UX perspective, mostly due to how Theia manages this (what does restarting Theia look like when it's running in your browser?). Without a way to tell the user what is happening, I expect most folks would think it's an issue with the plugin, and not a manual step they're required to do -- I only figured it out because I'm familiar with Theia and know where to look to diagnose an issue.

I could be convinced that this PR should be merged with future improvements down the line (and would *love* to have Scala support), but would defer to codeowners on the issue.  @ericwill WDYT?

ericwill:
> @olafurpg in terms of this PR, Che runs with multiple roots by default, so having to change the Theia folder is a very unintuitive flow. I would personally consider this a blocker from a UX perspective, mostly due to how Theia manages this (what does restarting Theia look like when it's running in your browser?). 

I'd be curious in knowing the answer to this question as well.

With respect to the subfolder issue, I don't think it's a big deal provided that the plug-in behaves the same way on VS Code (with a multi-root workspace). I am concerned about this though:

>     * The language server seems to fail to initialize on the first run (it has to download metals, and seems to log a series of errors afterwards)

@amisevsk do you still see this behaviour when trying the plugin?

amisevsk:
Here is a summary of the current flow:

1. Workspace starts, metals is downloaded
2. Metals doesn't work because of the issue described above; error
    ```
    ERROR:Remote plugin in vscode-metalso4t: promise rejection is not handled in two seconds: Error: Command metals-logs-toggle already exist
    ERROR:Remote plugin in vscode-metalso4t: promise rejection stack trace: Error: Command metals-logs-toggle already exist
        at CommandRegistryImpl.registerCommand (/remote-endpoint/node_modules/@theia/plugin-ext/lib/plugin/command-registry.js:99:19)
        at registerCommand (/remote-endpoint/node_modules/@theia/plugin-ext/lib/plugin/plugin-context.js:136:40)
        at Object.vscode.commands.registerCommand (/remote-endpoint/node_modules/@theia/plugin-ext-vscode/lib/node/plugin-vscode-init.js:45:20)
        at registerCommand (/tmp/vscode-unpacked/scalameta.metals.1.7.7.qptxwvjtyb.scalameta.metals-1.7.7.vsix/extension/out/extension.js:173:54)
        at vscode_1.commands.getCommands.then.values (/tmp/vscode-unpacked/scalameta.metals.1.7.7.qptxwvjtyb.scalameta.metals-1.7.7.vsix/extension/out/extension.js:305:25)
        at process._tickCallback (internal/process/next_tick.js:68:7)
    ```
    is logged but may be irrelevant.
3. You need to navigate the theia menus: `File > Open Workspace...` and open the `console-scala-simple` subfolder; this opens a new browser tab of Theia with the root folder opened; here, the plugin will initialize and download additional dependencies:
    ```
    $ du -h -d 2 /home/theia
    57.2M   /home/theia/.sbt/boot
    224.0K  /home/theia/.sbt/1.0
    57.4M   /home/theia/.sbt
    54.8M   /home/theia/.ivy2/cache
    54.8M   /home/theia/.ivy2
    269.6M  /home/theia/.cache/coursier
    12.4M   /home/theia/.cache/bloop
    8.0K    /home/theia/.cache/metals
    282.0M  /home/theia/.cache
    [...]
    394.3M  /home/theia
    ```
4. Once this process is done, you get a prompt from metals to import the build. Once this is done, everything works perfectly; I'm able to get completions, run and debug the main method without issue.
    - However, you get a popup in Theia saying 
        ```
        A new process is now listening on port 8212 but is listening on interface ::ffff:7f00:1 which is internal. You should change to be remotely available. Would you want to add a redirect for this port so it becomes available ?
        ```
        this isn't ideal but also isn't a huge deal

As I see it, we have the following issues:
- Dependencies have to download at start, and this can take *a while*. We should create a scala sidecar with these dependencies bundled rather than relying on `che-sidecar-java`. This does not block merging this PR, however, IMO.
- The number of mounts used in the PR is quite large; I'd prefer we not require 5 mounts, though this could potentially be mitigated by the sidecar described above (since it appears most mounts are for caching downloaded artifacts), and so I don't necessarily view this as a blocker either.
- The requirement of step 3 above is the biggest issue, since there's no notification that you need to do this -- a user trying to start this workspace will just see a nonfunctional plugin. Given that we don't have support for multi-root, and the plugin works quite well otherwise, I would lean to merging anyways, but would prefer some way of notifying the user of this limitation.
- The required download and the fact that we're storing it in a persistent volume could cause issues for us, as using 394.3MB of storage for each Scala workspace could exhaust your resources very quickly (on che.openshift.io, persistent volumes were 1Gi by default for a long time). This would also be helped by building the deps into a sidecar.

@olafurpg is there a convenient way to download these dependencies outside of running the plugin? If you can provide the steps to prepare all the dependencies used by metals beforehand, in a docker image, I'm happy to look into writing the sidecar dockerfile we'd need.

You can use the following devfile to test:

<details><summary>devfile</summary>
<p>

```yaml
metadata:
  name: scala-sbt-pr-check
projects:
  - name: console-scala-simple
    source:
      location: 'https://github.com/sideeffffect/console-scala-simple.git'
      type: git
      branch: master
components:
  - type: chePlugin
    reference: >-
      https://raw.githubusercontent.com/eclipse/che-plugin-registry/ee48be5ed73c5c8c6e69107d9296d5180b0c6378/v3/plugins/scalameta/metals/1.7.7/meta.yaml
  - mountSources: true
    memoryLimit: 512Mi
    type: dockerimage
    volumes:
      - name: sbt
        containerPath: /home/user/.sbt
      - name: ivy2
        containerPath: /home/user/.ivy2
      - name: coursier
        containerPath: /home/user/.cache/coursier
      - name: metals
        containerPath: /home/user/.cache/metals
      - name: bloop
        containerPath: /home/user/.cache/bloop
    alias: sbt
    image: 'quay.io/eclipse/che-java11-maven:nightly'
    env:
      - value: >-
          -J-XX:MaxRAMPercentage=50 -J-XX:+UseParallelGC
          -J-XX:MinHeapFreeRatio=10 -J-XX:MaxHeapFreeRatio=20
          -J-XX:GCTimeRatio=4 -J-XX:AdaptiveSizePolicyWeight=90
          -Dsun.zip.disableMemoryMapping=true -J-Xms20m
          -Djava.security.egd=file:/dev/./urandom -Duser.home=/home/user
        name: SBT_OPTS
      - value: >-
          -XX:MaxRAMPercentage=50 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
          -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4
          -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true
          -Xms20m -Djava.security.egd=file:/dev/./urandom
        name: JAVA_OPTS
      - value: >-
          -XX:MaxRAMPercentage=50 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
          -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4
          -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true
          -Xms20m -Djava.security.egd=file:/dev/./urandom
        name: JAVA_TOOL_OPTIONS
apiVersion: 1.0.0
commands:
  - name: sbt REPL
    actions:
      - workdir: '${CHE_PROJECTS_ROOT}/console-scala-simple'
        type: exec
        command: ./sbt
        component: sbt
  - name: sbt compile
    actions:
      - workdir: '${CHE_PROJECTS_ROOT}/console-scala-simple'
        type: exec
        command: ./sbt compile
        component: sbt
  - name: sbt test
    actions:
      - workdir: '${CHE_PROJECTS_ROOT}/console-scala-simple'
        type: exec
        command: ./sbt test
        component: sbt
  - name: sbt run
    actions:
      - workdir: '${CHE_PROJECTS_ROOT}/console-scala-simple'
        type: exec
        command: ./sbt run
        component: sbt
```

</p>
</details>

amisevsk:
I forgot to add: opening a subfolder in Theia in this way sort of breaks the navigation back to the dashboard: you still have the yellow arrow to open the sidebar but clicking it opens a tab with the original folder open.

tgodzik:
> @olafurpg is there a convenient way to download these dependencies outside of running the plugin? > If you can provide the steps to prepare all the dependencies used by metals beforehand, in a docker > image, I'm happy to look into writing the sidecar dockerfile we'd need.

You can use coursier for that: https://get-coursier.io/

Metals will not download anything a long it's available in the coursier cache (we do use it internally)

We do have a script in our codebase for fetching Metals itself:

https://github.com/scalameta/metals/blob/master/bin/test-release.sh

ericwill:
+1 for building a sidecar with the deps. Once we have that merged and built we can merge this PR with the correct sidecar container referenced in the meta.yaml file.

@benoitf  can you add a new repo to che-dockerfiles for scala? Something che-sidecar-scala sounds right.

benoitf:
@ericwill repo created

olafurpg:
The `test-release.sh` script downloads all publishable Metals artifacts but Metals also downloads on other published artifacts that it does not have a direct dependency on

* `ch.epfl.scala:bloop-frontend_2.12`: for the build server
* `org.scalameta:semanticdb-scala_{2.11.12,2.12.10,2.13.1}`: compiler plugin to power indexing. See https://github.com/scalameta/scalameta/blob/2ef927539dde5d3e53a403e260b6bd85b3898f92/bin/test-release.sh#L16-L21 
* `org.scalameta:scalafmt-core_2.12`: users customize their Scalafmt version via `.scalafmt.conf` so we would have to download a lot of versions if we want to always hit the cache
* `org.scalameta:mdoc_2.{11,12,12}`: to evaluate worksheets.

I don't think it's sustainable to track these artifacts in a bash script so it would probably be better to add a new main class or a subcommand to the `metals` binary that takes care of downloading all of these. Would that make sense?

olafurpg:
I opened a PR adding a new `DownloadDependencies` main class in the Metals module that does a best-effort attempt to pre-populate the Coursier caches, see https://github.com/scalameta/metals/pull/1501

You'll need a separate step to pre-populate the cache for an sbt/Maven/Gradle build.

You can launch this tool with the following script (requires Java 8 installation or newer), which should work on Linux and macOS.

```sh
METALS_VERSION=0.8.2 # Not released yet
curl -Lo coursier https://git.io/coursier-cli
chmod +x coursier
./coursier launch org.scalameta:metals_2.12:$METALS_VERSION --main scala.meta.metals.DownloadDependencies
```

olafurpg:
As for the multiroot workspace issue, we'll try to bump up the priority for that but we can't make any promises. Maybe we can come up with a solution that handles at least the situation for Theia (without dealing with more complicated scenarios like juggling multiple build server connections).

BTW, I really appreciate your help making this happen. I'm excited to see Metals become available to Theia and the Eclipse Che plugin registry!

tgodzik:
I am doing some similar prepopulation for sbt in https://github.com/tgodzik/metals-sample/blob/master/gitpod_init.sh, but the tool that @olafurpg added will be better for Metals and Bloop.

gattytto:
> Thank you very much @tgodzik for the input, I'll have to look into how other language plugins handle this, I admit I'm not familiar with the details there.

bazel extension behaves the same way, One needs to "add folder to workspace" with the project subfolder so it finds the WORKSPACE and other bazel related files.

sideeffffect:
Hello, is there something that can be done to merge this PR?

ericwill:
Yes, the sidecar container with the dependencies should be built. Please open a PR against https://github.com/che-dockerfiles/che-sidecar-scala on branch "1.7.7" and provide a Dockerfile that will house all the dependencies this plug-in needs to run. You can see examples of other such Dockerfiles in the [che-dockerfiles org](https://github.com/che-dockerfiles).

sideeffffect:
cool, I'll try to find some time this week to create the container ;)

sideeffffect:
I've switched to `che-sidecar-scala:1.8.6-7a4cdd9` sidecar container as per https://github.com/che-dockerfiles/che-sidecar-scala/pull/2#issuecomment-624295920
Is there anything else blocking this PR?

ericwill:
Is there a good sample project I can use to test this plugin?

amisevsk:
Played around a bit more and was able to import the project using the metals tab in the sidebar, at which point the lang server works.

soronpo:
Maybe updating to the latest Metals (0.9.0) will improve some of the issues.

sideeffffect:
https://github.com/sideeffffect/console-scala-simple
(created for the devfile at https://github.com/eclipse/che-devfile-registry/pull/186 )

tgodzik:
> Metals still has to download on workspace start, this can take a bit

Most things should be already downloaded when using the downloader provided by @olafurpg so maybe we need to add something more? Or maybe the cache is not actually populated?

> The prompt that the workspace is downloading metals appears every time, even if it's already downloaded

That should be fixed in 1.9.0

> Once actual project is opened, sbt bloopinstall takes a few minutes
> Once everything is installed/downloaded, plugin fails to initialize it seems:
> error message: Import project failed, no functionality will work. See the logs for more details

That unfortunately is needed as each workspace can contain different dependencies and we need to download them. It seems it failed in this instance:
`INFO  skipping build import with status 'Failed'`

sideeffffect:
upgraded to `1.9.0` (Metals 0.9.0).
Repeated downloading of Metals is most likely because the sidecar contained it under `/root` directory (instead of `/home/theia`). Fix is on its way: https://github.com/che-dockerfiles/che-sidecar-scala/pull/3
Otherwise I tested it, and it the plugin seems to work fine (you have to open the workspace containing an sbt project, of course).

amisevsk:
In case anyone else wants to test out the changes, here's the devfile I used for my last review: https://gist.github.com/amisevsk/e63d09341701c1d54902887c6cabac17

If you want to create a redhat developer account, you can start up the workspace using

https://che.openshift.io/f?url=https://gist.githubusercontent.com/amisevsk/e63d09341701c1d54902887c6cabac17/raw/271602f91ba693604cd98d6fe89fc9923d5eed8a/scala.devfile.yaml

If the meta.yaml here is updated (i.e. once https://github.com/che-dockerfiles/che-sidecar-scala/pull/3 is merged), you just need to update [line 13](https://gist.github.com/amisevsk/e63d09341701c1d54902887c6cabac17#file-scala-devfile-yaml-L12-L13) of the devfile to point at the new raw meta.yaml.

sideeffffect:
The sidecar and the plugin have been updated :tada:

olafurpg:
> The prompt that the workspace is downloading metals appears every time, even if it's already downloaded

This was fixed in the latest release (it was a regression)

soronpo:
Pardon my ignorance, but what does this mean for eclipse che exactly? Do we have metals in the market place now?

ericwill:
> Pardon my ignorance, but what does this mean for eclipse che exactly? Do we have metals in the market place now?

If by market place you mean the Che plugin registry, yes. After the next Che release you will see the Scala plugin in the available plugins list inside Che.

soronpo:
Yeah, sorry, the market place is from VS Code. 
So the next release that will include Metals is 7.13? https://github.com/eclipse/che/milestone/116

ericwill:
Correct!

PR #362, opened 2020-01-29
chore(github): Code Owners should match the team

benoitf:
### What does this PR do?

Based on last che SOS and based on repositories scope:

https://docs.google.com/document/d/1h6t4novpcRHjxe8DrEhTamDmit9VviOqcMQNtahM9uE/edit#heading=h.nh4qax6oermn

owners of che-plugin-registry is Eric with plugins team

Change-Id: I286e83ced5a996974095b07e85ff2c5229dd9a26

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 # Global Owners
-* @vparfonov @l0rd @rhopp @skabashnyuk @amisevsk @nickboldt @ibuziuk
+* @ericwill @vinokurig @vitaliy-guliy @benoitf
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 # Global Owners
-* @ericwill @vinokurig @vitaliy-guliy @benoitf
+* @ericwill @vinokurig @vitaliy-guliy @benoitf @amisevsk @nickboldt


benoitf:
@nickboldt reviews are always welcomed but owners should be only Plugin team.

l0rd:
@nickboldt you will be able to merge/review PRs even if you are not an owner.  The only difference is that you won't be automatically added as a reviewer to every PR anymore.

benoitf:
I think that some repos have a setting that is only allowing to merge for code owners only

benoitf:
why is it been modified before merging ?

We agreed on code owners, here it has been changes

If I haven't merge it yet I was waiting to allow non code owners to lerge

nickboldt:
There was some disagreement about the value of removing me and Angel, and no one from the new Lang team objected so I fixed the PR and merged it. :)

benoitf:
@nickboldt about disagreement this is why it has been discussed on the two previous calls.
it has been decided that owners are the one responsible of the repository and that other committers should still be able to merge.

Eric is now an Eclipse committer so there is no issue and there were also other a couple of backups (his team)

> If you want to instead fix the repo to allow non-codeowners to merge, then there's little point in having a codeowners file, in which case I'd have the same permissions you plan to bestow on my by making the repo config change.

No, it's because owners are notified are are added automatically as reviewer.
BTW this is what has been decided on the calls, so I would avoid to discuss the same things again on PR comments.

PR #368, opened 2020-01-30
Add notion about building vsix extensions for plugin registry

mkuznyetsov:
Signed-off-by: Mykhailo Kuznietsov <mkuznets@redhat.com>

### What does this PR do?
Add notion about building vsix extensions for plugin registry

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,9 @@
 
 # Eclipse Che plugin registry
 
+## Building and publishing third party VSIX extensions for plugin registry
+See: https://github.com/redhat-developer/codeready-workspaces/blob/master/devdoc/building/build-vsix-extension.adoc
+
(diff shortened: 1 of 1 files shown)


nickboldt:
there used to be a way to link job results on internal jenkins to a public facing one but i suspect that died when we moved from the bos.mw lab to the central ci psi/openstack thing.

THAT SAID we could consider moving / copying the job onto centos.ci but we'd need to implement a way for the tools@ internal user to push from an external machine into RH to be published on dl.jb.org. OR... we could instead implement the build process so it publishes to another location, like a GH repo.

nickboldt:
wdyt @mkuznyetsov ? Something to explore after 2.1.0 is done?

PR #370, opened 2020-02-03
Updated vscode-xml to 0.10.1

JPinkney:
Signed-off-by: Josh Pinkney <joshpinkney@gmail.com>

### What does this PR do?
This PR updates vscode-xml to 0.10.1 as requested in https://github.com/eclipse/che/issues/15592


nickboldt:
does this deprecate/replace previous version? if so you can set the deprecation directive in 0.9.1

PR #373, opened 2020-02-05
chore(github): Code Owners should match the team

benoitf:
### What does this PR do?
To start, owners are the one which are responsible of the repository.
Other committers will be able to merge as well when repository settings will be updated (as discussed in the Eclipse Che calls)

Change-Id: Ia89283f128d1cf39a5b5ec6f295b783a82404deb


nickboldt:
> when repository settings will be update

When is that going to happen and who has permission to make the change?

benoitf:
it's already updated

PR #375, opened 2020-02-05
Add Vale to the Che Plugin Registry

ericwill:
Initial contribution of the vscode-vale extension to the Che Plugin Registry.
The sidecar container for this extension contains a basic Vale config file.

See eclipse/che#15856


ericwill:
> Testing the plugin with the [che-docs devfile](https://raw.githubusercontent.com/eclipse/che-docs/master/devfile.yaml) I keep seeing the error toast notification:
> 
> ```
> SyntaxError: Unexpected token a in JSON at position 0
> ```

Interesting issue -- it seems that vale expects asciidoctor to be present when linting AsciiDoc files. When it doesn't find asciidoctor, it throws an "asciidoctor missing!" message into stdout, instead of the expected JSON. The first letter "a" of this error message is being picked up by vscode-vale, which in turn throws an error because it's not valid JSON.

I've updated the sidecar image and linting AsciiDoc files now works as expected.

amisevsk:
cc: @themr0c

PR #376, opened 2020-02-06
Update Java extensions to latest

tsmaeder:
### What does this PR do?
Updates Java related VS Code extensions to latest versions

Fix for: https://github.com/eclipse/che/issues/15560


v3/plugins/redhat/java8/0.55.1/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - image: "quay.io/eclipse/che-sidecar-java:8-0cfbacb"
+      name: vscode-java
+      memoryLimit: "1500Mi"
+      volumes:
+      - mountPath: "/home/theia/.m2"
+        name: m2
+  extensions:
+    - https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-java-debug/vscode-java-debug-0.24.0.vsix
+    - https://download.jboss.org/jbosstools/static/jdt.ls/stable/java-0.55.1-1984.vsix
+
```
amisevsk: Please add a trailing newline in case some of the build scripts expect it.

nickboldt:
Is 0.55 the version we want in CRW 2.1? @tsmaeder @fbricon

tsmaeder:
> +1, but I might prefer to see `1.2.0-0.55.1` than `1.2.0a` as the new version of quarkus-java8, as it's more meaningful / explains the re-release.
> 
Yeah, but that does not really scale, as we have multiple extensions in there. Using the VS Code extensions version as the Che plugin version only really works when you have a single extension. If what we have inside the Che plugin is relevant, we need a better way to show it.

v3/plugins/redhat/java8/0.55.1/meta.yaml
tsmaeder: Requiring an _empty_ newline at the end of the file is so stupid it makes me want to scream. But yeah, fair point: I'll bow to the great machine

nickboldt:
> Yeah, but that does not really scale

True dat. The flawed system is flawed. :(

v3/plugins/redhat/java8/0.55.1/meta.yaml
nickboldt: https://i.redd.it/zibdmw9m66141.jpg You're not the only one annoyed by syntactic whitespace :)

PR #377, opened 2020-02-12
Add new Quarkus plugin version

mkuznyetsov:
Signed-off-by: Mykhailo Kuznietsov <mkuznets@redhat.com>

### What does this PR do?
Add new version of quarkus plugin, deprecate the older one


nickboldt:
note in downstream for quarkus-java11 1.3  we use newer extensions:

https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-plugin-registry/v3/plugins/redhat/quarkus-java11/1.3.0/meta.yaml#L26

So should do the same in upstream IMHO.

PR #378, opened 2020-02-13
Secure the che-machine-exec and che-theia plugins

metlos:
### What does this PR do?

Modify the che-machine-exec and che-theia plugins to only bind to 127.0.0.1.

This is in support eclipse/che#15651 and depends on eclipse/che#15890 and https://github.com/eclipse/che-theia/pull/626 both being merged.

### DIFF
--- a/v3/plugins/eclipse/che-machine-exec-plugin/nightly/meta.yaml
+++ b/v3/plugins/eclipse/che-machine-exec-plugin/nightly/meta.yaml
@@ -27,3 +27,5 @@ spec:
      image: "quay.io/eclipse/che-machine-exec:nightly"
      ports:
        - exposedPort: 4444
+     command: ['/go/bin/che-machine-exec', '--static', '/cloud-shell', '--url', '127.0.0.1:4444']
+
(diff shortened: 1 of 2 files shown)


v3/plugins/eclipse/che-machine-exec-plugin/nightly/meta.yaml
```diff
@@ -27,3 +27,5 @@ spec:
      image: "quay.io/eclipse/che-machine-exec:nightly"
      ports:
        - exposedPort: 4444
+     command: ['/go/bin/che-machine-exec', '--static', '/cloud-shell', '--url', '127.0.0.1:4444']
```
sleshchenko: @metlos Does it influence Single-user Che where we don't run jwtproxy?

sleshchenko: Seems it breaks Single-User Che https://github.com/eclipse/che/issues/16053

AndrienkoAleksandr: I think @metlos in PTO.

PR #379, opened 2020-02-13
Add option to use digests for images referenced in registry

amisevsk:
### What does this PR do?
Add script /build/scripts/write_image_digests.sh, which will rewrite all image references in registry to use the current digest for each image tag. E.g.
```
  containers:
    - image: "quay.io/eclipse/che-sidecar-java:8-0cfbacb"
```
is replaced with
```
  containers:
    - image: "quay.io/eclipse/che-sidecar-java@sha256:4fb35b4e112a05c82353ad0e1f5eb35d1ba204be475b72f3c753b0e26c9e668d" # tag: quay.io/eclipse/che-sidecar-java:8-0cfbacb
```
To enable this functionality, it is necessary to install `skopeo` in the builder images, which required changing adding the CentOS8-AppStream repo for the rhel build.

### Additional info
Digest-rewriting functionality is disabled by default and can be enabled by using the `./build.sh` option `--use-digests`, or by passing docker build arg `USE_DIGESTS=true`. 

Existing airgap options are unaffected (you can still set env vars to override registry, organization, and tag). In the case of overriding tags, image digests are replaced with the tag specified.

There's also a commit included to update all copyright years for 2020.

### Testing
Image is available as `amisevsk/che-plugin-registry:digests`. I've tested the changes on the dev cluster and locally.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (c) 2012-2018 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
 # This program and the accompanying materials are made
 # available under the terms of the Eclipse Public License 2.0
 # which is available at https://www.eclipse.org/legal/epl-2.0/
--- a/build.sh
+++ b/build.sh
@@ -14,7 +14,7 @@ REGISTRY="quay.io"
 ORGANIZATION="eclipse"
 TAG="nightly"
 LATEST_ONLY=false
-USE_DIGESTS=true
+USE_DIGESTS=false
 OFFLINE=false
 DOCKERFILE="./build/dockerfiles/Dockerfile"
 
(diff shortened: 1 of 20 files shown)


davidfestal:
Is it possible, for now, to have that as an opt-in mechanism ? With a 'use-digest' option that would be disabled on the build by default ?

davidfestal:
Then we could use it on the CRW build at first, and then generalize it to Upstream in a second step...

amisevsk:
@davidfestal done.

davidfestal:
Thanks Angel !

nickboldt:
new commit looks good, thanks

davidfestal:
According to @l0rd  comment in the sibling PR: https://github.com/eclipse/che-devfile-registry/pull/195#issuecomment-586448303, we should be OK to merge this one as well/

PR #381, opened 2020-02-18
Add CPU limit/request and RAM request into meta.yaml schema

mshaposhnik:
### What does this PR do?
Adds CPU limit/request and RAM request into meta.yaml schema

Ticket: https://github.com/eclipse/che/issues/13709


mshaposhnik:
ci-build

build/scripts/meta.yaml.schema
```diff
@@ -120,6 +120,18 @@
                 "type": "string",
                 "pattern": "[1-9][0-9]*([Gg]i|(Mm)i|(Kk)i|b)?"
               },
+              "memoryRequest": {
+                 "type": "string",
+                 "pattern": "[1-9][0-9]*([Gg]i|(Mm)i|(Kk)i|b)?"
```
amisevsk: This regexp doesn't seem to be accurate (but I see it's copied from the limit above). This will match any number not starting with 0, and ignore the suffix completely in all cases except `Gi`.

build/scripts/meta.yaml.schema
```diff
@@ -120,6 +120,18 @@
...
+              },
+              "cpuLimit": {
+                 "type": "string",
+                 "pattern": "(^[0-9]+(.[0-9][0-9]*)?$|^[1-9][0-9]+(m)?$)"
```
amisevsk:
```suggestion
                 "pattern": "(^[0-9]+(.[0-9]+)?$|^[1-9][0-9]*(m)$)"
```

build/scripts/meta.yaml.schema
mshaposhnik: FIxed it

build/scripts/meta.yaml.schema
mshaposhnik: Thanks

README.md
```diff
@@ -111,7 +111,10 @@ spec:                  # spec (used to be che-plugin.yaml)
   containers:          # optional; sidecar containers for plugin
     - image:
       name:              # name used for sidecar container
-      memorylimit:       # Kubernetes/OpenShift-spec memory limit string (e.g. "512Mi")
+      memoryLimit:       # Kubernetes/OpenShift-spec memory limit string (e.g. "512Mi")
+      memoryRequest:     # Kubernetes/OpenShift-spec memory request string (e.g. "256Mi")
+      cpuLimit:          # Kubernetes/OpenShift-spec CPU limit string (e.g. "500m")
```
sleshchenko: Consider referencing K8s documentation here https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-cpu

build/scripts/meta.yaml.schema
```diff
@@ -118,7 +118,19 @@
...
-                "pattern": "[1-9][0-9]*([Gg]i|(Mm)i|(Kk)i|b)?"
+                "pattern": "^[1-9][0-9]*([Gg]i?|[Mm]i?|[Kk]i?|b)?$"
+              },
+              "memoryRequest": {
+                 "type": "string",
+                 "pattern": "^[1-9][0-9]*([Gg]i?|[Mm]i?|[Kk]i?|b)?$"
+              },
+              "cpuLimit": {
+                 "type": "string",
+                 "pattern": "(^[0-9]+(.[0-9]+)?$|^[1-9][0-9]*(m)$)"
```
sleshchenko: 1. dot means any character, so 2x2 is a valid cpuLimit )
2. Brackets at the beginning and the end are redundant
3. Do you really need a separate group for `m` char?
```suggestion
                 "pattern": "^[0-9]+(\\.[0-9]+)?$|^[1-9][0-9]*m$"
```
Also this expression does not cover requirement 
> precision finer than 1m is not allowed. For this reason, the form 100m might be preferred
Up to you to cover it or not

build/scripts/meta.yaml.schema
```diff
@@ -118,7 +118,19 @@
...
+              },
+              "cpuRequest": {
+                 "type": "string",
+                 "pattern": "(^[0-9]+(.[0-9]+)?$|^[1-9][0-9]*(m)$)"
```
sleshchenko:
```suggestion
                 "pattern": "^[0-9]+(\\.[0-9]+)?$|^[1-9][0-9]*m$"
```

README.md
```diff
@@ -111,7 +111,10 @@ spec:                  # spec (used to be che-plugin.yaml)
   containers:          # optional; sidecar containers for plugin
     - image:
       name:              # name used for sidecar container
-      memorylimit:       # Kubernetes/OpenShift-spec memory limit string (e.g. "512Mi")
+      memoryLimit:       # Kubernetes/OpenShift-spec memory limit string (e.g. "512Mi")
```
sleshchenko: Consider adding a reference to k8s docs https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-memory

build/scripts/meta.yaml.schema
```diff
@@ -118,7 +118,19 @@
               },
               "memoryLimit": {
                 "type": "string",
-                "pattern": "[1-9][0-9]*([Gg]i|(Mm)i|(Kk)i|b)?"
+                "pattern": "^[1-9][0-9]*([Gg]i?|[Mm]i?|[Kk]i?|b)?$"
```
sleshchenko: Why don't you believe that some plugin may need 1Ti of memory :smile:

sleshchenko: Good catch with `^` and `$`. It makes sense to review all fields where we use regexps, like `image`.

build/scripts/meta.yaml.schema
sleshchenko: Seems the same changes are required for `initContainers`

mshaposhnik: About `initContainers`.. do we need some code changes in Che server ho handle that ?

sleshchenko: I don't know

build/scripts/meta.yaml.schema
```diff
@@ -118,7 +118,19 @@
...
+              },
+              "memoryRequest": {
+                 "type": "string",
+                 "pattern": "^[1-9][0-9]*([Gg]i?|[Mm]i?|[Kk]i?|b)?$"
+              },
+              "cpuLimit": {
+                 "type": "string",
+                 "pattern": "^[0-9]+(\\.[0-9]+)?$|^[1-9][0-9]*m$"
```
sleshchenko:
```suggestion
                 "pattern": "^[1-9]+[0-9]*(\\.[0-9]+)?$|^[1-9][0-9]*m$"
```
consider to prevent numbers like `000003`

PR #387, opened 2020-02-20
Add new version of VSCode Java - 0.57.0

svor:
Signed-off-by: svor <vsvydenk@redhat.com>

### What does this PR do?
Adds new version of VSCode Java - 0.57.0

### Related issue:
https://github.com/eclipse/che/issues/16051


nickboldt:
I've merged this, but I think it's incomplete - doesn't solve two problems:

* 0.55.1 should be marked deprecated and force everyone to update to 0.57.0
* quarkus plugin relies on 0.55.1 so I'm assuming that should ALSO be updated w/ deprecation path to 0.57.0

ericwill:
LGTM but Nick already merged. :+1:

nickboldt:
Yes... But see follow-up.

Should we do that too, as a blocker for 7.9.0?

ericwill:
> Yes... But see follow-up.
> 
> Should we do that too, as a blocker for 7.9.0?

Is it mandatory we deprecate 0.55.1 for 7.9.0? Or can we leave it for 7.10.0? IMO it's best to only change what's absolutely necessary, and do more thorough testing for Quarkus on 0.57.0 to make sure nothing breaks there.

v3/plugins/redhat/java11/0.57.0/meta.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v2
+publisher: redhat
+name: java
```
nickboldt: should this be java11 ?

v3/plugins/redhat/java8/0.57.0/meta.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v2
+publisher: redhat
+name: java
```
nickboldt: should this be java8 ?

svor:
@nickboldt it should be 
https://github.com/eclipse/che-plugin-registry/pull/391

nickboldt:
thanks, fixed in 7.9.x (and 7.9.0 rebuilt); merged in master too.

nickboldt:
followup work in https://github.com/eclipse/che/issues/16107

PR #390, opened 2020-02-21
fixup name fields in java11 and...

nickboldt:
fixup name fields in java11 and java8

Change-Id: I5e60e259f8d258095111089fa40a5fe774023414


nickboldt:
Should have a startup test / CI test that fails if this happens. cc: @amisevsk

PR #393, opened 2020-02-22
Fix execution of image build checks

mkuznyetsov:
### What does this PR do?
Fix check scripts execution flow


nickboldt:
I was wondering about that shorter syntax. So... did not execute as expected?

mkuznyetsov:
@nickboldt the shorter syntax before meant that if any of the scripts would fail (which they did), it would just continue executing after `||` operator

@amisevsk yes, I think too separating lines into multiple run commands would make the logic simpler, I'll do that (and same for rhel.Dockerfile)

nickboldt:
IMHO chained commands with `&&` vs separate RUN commands behave the same way (first one that fails breaks the chain) but if you guys prefer this for debugging, so be it.

PR #394, opened 2020-02-24
minor tweaks from downstream

nickboldt:
minor tweaks from downstream


build/dockerfiles/rhel.install.sh
```diff
@@ -11,7 +11,10 @@
 microdnf install -y findutils bash wget yum gzip tar jq python3-six python3-pip skopeo && microdnf -y clean all && \
 # install yq (depends on jq and pyyaml - if jq and pyyaml not already installed, this will try to compile it)
 if [[ -f /tmp/root-local.tgz ]] || [[ ${BOOTSTRAP} == "true" ]]; then \
-    mkdir -p /root/.local; tar xf /tmp/root-local.tgz -C /root/.local/; rm -fr /tmp/root-local.tgz;  \
+    mkdir -p /root/.local; \
```
amisevsk: Still think the `\` are unnecessary here, but I won't argue.

nickboldt: you're not wrong... but thanks. :D

PR #399, opened 2020-02-27
Add script to automatically detect and file issues for out of date plugins

ericwill:
This PR contributes a Python script that:
	-reads all plugins in a plugin registry
	-checks their versions
	-checks the upstream repositories of their vsix files
	-checks whether the plugin's vsix file needs to be updated
	-prints out a list of plugins that need to be updated

Additionally, this script can also file GitHub issues for any plugin
that needs to be updated. A README with installation/usage instructions
has been provided.

Fixes eclipse/che#16215


tools/README.md
```diff
@@ -0,0 +1,38 @@
...
+This directory contains a set of tools to facilitate certain tasks that a Che developer may encounter.
+
+## Plugin Updates
+
+All Che plugins depend on VS Code extensions, and these are usually hosted on other repositories. Sometimes these extensions can be out of date,
+as a newer version of the extension has been released upstream. In order to maintain the Che Plugin Registry and keep it up to date, a script
+has been provided which can check for updates, report which plugins need updating, and automatically file issues for those plugins that do
+need updating.
+
+Installation: create a virtual environment with the following Python packages installed
```
benoitf: Should you provide Dockerfile with the required runtime for this script ?

tools/plugin_checker.py
```diff
@@ -0,0 +1,188 @@
...
+# Add arguments
+parser = argparse.ArgumentParser(description = DESC_TEXT)
+parser.add_argument("-f", "--file", help="file issues against the GitHub repo", action="store_true")
+parser.add_argument("-s", "--silent", help="run this script silently without any report details", action="store_true")
+args = parser.parse_args()
+
+
+# NOTE: edit the fields below to customize where and how the script files issues
+# Access token generated by GitHub
+TOKEN = ''
```
benoitf: could it work with environnent variable instead of editing the script ?

tools/plugin_checker.py
```diff
@@ -0,0 +1,188 @@
...
+        else:
+            PLUGINS_FAILED_TO_CHECK = PLUGINS_FAILED_TO_CHECK + k + "\n"
+    
+    # Update report messages
+    PLUGINS_TO_UPDATE = PLUGINS_TO_UPDATE + "\n"
+    PLUGINS_FAILED_TO_CHECK = PLUGINS_FAILED_TO_CHECK + "\n"
+    PLUGINS_UP_TO_DATE = PLUGINS_UP_TO_DATE + "\n"
+
+def compare_versions(base_version, compared_version):
+    '''Return True iff base_version is less than compared_version.'''
```
benoitf:
```suggestion
    '''Return True if base_version is less than compared_version.'''   
```

tools/README.md
ericwill: Sure, that's an interesting idea!

tools/plugin_checker.py
ericwill: This isn't a typo, it's meant to signify ["if and only if"](https://en.wikipedia.org/wiki/If_and_only_if#Origin_of_iff_and_pronunciation)

tools/plugin_checker.py
ericwill: Yes, I'll investigate it.

tools/README.md
ericwill: @benoitf just to clarify, what is your intended use case? Someone would be able to build the container and run the script from in there? Should the token/repository details be editable in the Dockerfile, or as environment variables when running the script?

tools/plugin_checker.py
benoitf: ok might be better to fully write the word

tools/README.md
benoitf: use case was to run the script packaged into a container and by just providing env variable (so the container can run everywhere and doesn't contain any token/secrets) I'm able to use it

ericwill:
> no objection as first run, most comments are improvements

I don't mind making the improvements today

tools/plugin_checker/plugin_checker.py
```diff
@@ -0,0 +1,200 @@
+#! /usr/bin/env python
+
```
benoitf: Missing copyrights

tools/plugin_checker/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+FROM python:3.8.2-slim
+
+RUN apt-get update && \
+    apt-get install wget -y && \
+    wget -O - https://deb.nodesource.com/setup_10.x | bash - && \
```
benoitf: curious. Why do we need nodejs runtime ?

benoitf:
FYI there are some indent issues
https://pythonbuddy.com/

tools/plugin_checker/Dockerfile
ericwill: Oops, leftover from copy-paste. I'll remove it.

tools/plugin_checker/plugin_checker.py
```diff
@@ -0,0 +1,208 @@
...
+    sys.exit()
+
+find_plugin_versions()
+check_plugins_for_updates()
+clean_up()
+
+if not args.silent:
+    print("\n" + PLUGINS_TO_UPDATE)
+    print(PLUGINS_UP_TO_DATE)
+    print(PLUGINS_FAILED_TO_CHECK)
```
amisevsk: Missing newline

ericwill: At the end of the file or in the print statement?

JPinkney: End of the file, if you go into the `files changed` view it shows a little icon saying no newline but for some reason, it doesn't appear here

ericwill: Done!

PR #302, opened 2019-11-26
[plugin registry] remove support for node <...

nickboldt:
[plugin registry] remove support for node < 8 ( #14927 )

Change-Id: I3572d0c9d42996f5048269e2d98a5ec1ac84960c
Signed-off-by: nickboldt <nboldt@redhat.com>

PR #400, opened 2020-02-28
Add node-debug VS Code extensions to Che Typescript plugin

azatsarynnyy:

### What does this PR do?

Adds VS Code node-debug extensions to Che Typescript plugin as `@theia/debug-nodejs` Theia extension is removed in upstream Theia.

Devfile to check
```
metadata:
  name: pr-che_theia-643-t1v9q
attributes:
  persistVolumes: 'false'
components:
  - type: cheEditor
    reference: >-
      https://raw.githubusercontent.com/chepullreq4/pr-check-files/master/che-theia/pr-643/simple/che-theia-editor.yaml
  - type: chePlugin
    reference: >-
      https://raw.githubusercontent.com/eclipse/che-plugin-registry/8abc243e3392b6181eb8914ca1c7705c15cec3e5/v3/plugins/che-incubator/typescript/1.35.1/meta.yaml
apiVersion: 1.0.0

```

### DIFF
--- a/v3/plugins/che-incubator/typescript/1.35.1/meta.yaml
+++ b/v3/plugins/che-incubator/typescript/1.35.1/meta.yaml
@@ -17,3 +17,5 @@ spec:
     memoryLimit: '512Mi'
   extensions:
   - https://download.jboss.org/jbosstools/vscode/3rdparty/ms-code.typescript/che-typescript-language-1.35.1.vsix
+  - https://github.com/theia-ide/vscode-node-debug/releases/download/v1.35.3/node-debug-1.35.3.vsix
+  - https://github.com/theia-ide/vscode-node-debug2/releases/download/v1.33.0/node-debug2-1.33.0.vsix


azatsarynnyy:
> Should we build these from source as well, or is pointing directly to the theia-ide GitHub allowed?

@ericwill No, I guess. But I'm not sure.
cc @benoitf @nickboldt

nickboldt:
The rules:

[Product] if we need the .vsix in CRW, we MUST build from source and publish binary & source.tgz to the same location (eg., a github release w/ BOTH sources and binary or dl.jboss.org)

[Project] if we need it in Che and the 3rd party vendor is not publishing sources+binary in the same place (eg., the binary is only in MS VSCode Mktplace) then we need to rebuild it as we cannot fetch from microsoft.com site to use in our builds (TOS violation)

[Community] if we want to be Good Open Source Consumers and Compliant With The Open Source Licenses, then **anything we redistribute** in Che or CRW must exist in a form where the **binary and sources exist in the same place** (eg., same GH release or a folder on dl.jb.org). The less-friendly approach is to know the sources exist, but make it a few steps harder for someone to find them and match the exact version used to the binary we redistribute in Che / CRW.

v3/plugins/che-incubator/typescript/1.35.1/meta.yaml
```diff
@@ -17,3 +17,5 @@ spec:
     memoryLimit: '512Mi'
   extensions:
   - https://download.jboss.org/jbosstools/vscode/3rdparty/ms-code.typescript/che-typescript-language-1.35.1.vsix
+  - https://github.com/theia-ide/vscode-node-debug/releases/download/v1.35.3/node-debug-1.35.3.vsix
```
nickboldt: no-debug is deprecated and only needed for Node < 8. 

https://github.com/eclipse/che/issues/14927 
https://github.com/eclipse/che-plugin-registry/pull/302

We removed it from Che. Please don't add it back.

azatsarynnyy:
I'm closing the PR as we already have [Node Debug Che Plugin](https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/ms-vscode/node-debug2/1.33.0/meta.yaml)
Thanks @nickboldt for the detailed explanation! :+1:

PR #417, opened 2020-03-04
Modify the che-machine-exec and che-theia plugins to only bind to127.0.0.1.

metlos:
### What does this PR do?
This re-introduces the changes done in #378 

This PR depends on eclipse/che#16227 be merged to Che server master.

### DIFF
--- a/v3/plugins/eclipse/che-machine-exec-plugin/nightly/meta.yaml
+++ b/v3/plugins/eclipse/che-machine-exec-plugin/nightly/meta.yaml
@@ -27,3 +27,5 @@ spec:
      image: "quay.io/eclipse/che-machine-exec:nightly"
      ports:
        - exposedPort: 4444
+     command: ['/go/bin/che-machine-exec', '--static', '/cloud-shell', '--url', '127.0.0.1:4444']
+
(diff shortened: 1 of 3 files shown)


metlos:
> Why are we setting this value here and not in https://github.com/eclipse/che-machine-exec/, etc.?

For example the integration tests of Che Theia require it to be exposed on a "public" IP address. Che machine exec has a configurable host it listens on. IMHO it would be weird to hardcode the default host to 127.0.0.1, because that is not something you generally do. Hence, the next place "up the chain" is meta.yaml where we can secure the plugin by invoking it properly.

v3/plugins/eclipse/che-machine-exec-plugin/nightly/meta.yaml
```diff
@@ -27,3 +27,5 @@ spec:
      image: "quay.io/eclipse/che-machine-exec:nightly"
      ports:
        - exposedPort: 4444
+     command: ['/go/bin/che-machine-exec', '--static', '/cloud-shell', '--url', '127.0.0.1:4444']
+
```
sleshchenko:
```suggestion
```

v3/plugins/eclipse/che-machine-exec-plugin/nightly/meta.yaml
```diff
@@ -27,3 +27,5 @@ spec:
      image: "quay.io/eclipse/che-machine-exec:nightly"
      ports:
        - exposedPort: 4444
+     command: ['/go/bin/che-machine-exec', '--static', '/cloud-shell', '--url', '127.0.0.1:4444']
```
sleshchenko: Do you plan to update cloud-shell as well where `quay.io/eclipse/che-machine-exec:nightly` is used
https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/eclipse/cloud-shell/nightly/meta.yaml#L26 ?

PR #418, opened 2020-03-04
WIP: Add script that allows you to run tests against plugins in PRs

JPinkney:
### What does this PR do?

This PR adds in the ability to run language tests against incoming PRs.

It is currently hardcoded to run against a previous PR for testing purposes

PR #420, opened 2020-03-06
Don't rewrite image digests for images we can't read

amisevsk:
### What does this PR do?
Don't overwrite image references with nonsense when we can't access the image being used.

Fixes https://github.com/eclipse/che/issues/16281

### DIFF
--- a/build/scripts/write_image_digests.sh
+++ b/build/scripts/write_image_digests.sh
@@ -8,14 +8,29 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
+LOG_FILE="/tmp/image_digests.log"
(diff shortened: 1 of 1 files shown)


amisevsk:
Looks like CI is borked, given that this PR doesn't actually change plugins (and digests is [disabled](https://github.com/eclipse/che-plugin-registry/blob/master/build/dockerfiles/Dockerfile#L18) by default) I'm merging.

JPinkney:
@amisevsk https://github.com/eclipse/che-plugin-registry/pull/418 is still WIP thats why, just ignore anything that comes from that PR check for now

PR #424, opened 2020-03-10
Use PNG icons for Kubernetes plugin

vitaliy-guliy:

### What does this PR do?

- Allows to use PNG images for plugins
- Sets proper images for Kubernetes plugin

Solves issue https://github.com/eclipse/che/issues/14502

### DIFF
--- a/build/scripts/meta.yaml.schema
+++ b/build/scripts/meta.yaml.schema
@@ -34,7 +34,7 @@
     },
     "icon": {
       "type": "string",
-      "pattern": "[-._a-zA-Z0-9]+(.svg|.png)[^.]*$"
+      "pattern": "[-._a-zA-Z0-9]+?\\.(?:svg|png)(?:\\?.*)?$"
     },
     "description": {
       "type": "string"
--- a/build/scripts/meta.yaml.schema
+++ b/build/scripts/meta.yaml.schema
@@ -34,7 +34,7 @@
     },
     "icon": {
       "type": "string",
-      "pattern": "[-._a-zA-Z0-9]+.svg[^.]*$"
+      "pattern": "[-._a-zA-Z0-9]+(.svg|.png)[^.]*$"
     },
     "description": {
       "type": "string"
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ type:                  # plugin type; e.g. "Theia plugin", "Che Editor"
 displayName:           # name shown in user dashboard
 title:                 # plugin title
 description:           # short description of plugin's purpose
-icon:                  # link to SVG icon
+icon:                  # link to SVG or PNG icon
 repository:            # URL for plugin (e.g. Github repo)
 category:              # see [1]
 firstPublicationDate:  # optional; see [2]
--- a/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17/meta.yaml
+++ b/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17/meta.yaml
@@ -6,7 +6,7 @@ type: VS Code extension
 displayName: Kubernetes
 title: Kubernetes Tools
 description: Develop, deploy and debug Kubernetes applications
-icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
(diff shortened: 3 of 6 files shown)


build/scripts/meta.yaml.schema
```diff
@@ -34,7 +34,7 @@
     },
     "icon": {
       "type": "string",
-      "pattern": "[-._a-zA-Z0-9]+.svg[^.]*$"
+      "pattern": "[-._a-zA-Z0-9]+(.svg|.png)[^.]*$"
```
ericwill: The periods in the filename need to be escaped, otherwise this pattern will match names like `filepng` or `filesvg`.

`(.svg|.png)` -> `(\.svg|\.png)`

vitaliy-guliy: Updated with following
`[-._a-zA-Z0-9]+?\.(?:svg|png)(?:\?.*)?$`

build/scripts/meta.yaml.schema
```diff
@@ -34,7 +34,7 @@
     },
     "icon": {
       "type": "string",
-      "pattern": "[-._a-zA-Z0-9]+.svg[^.]*$"
+      "pattern": "[-._a-zA-Z0-9]+?\\.(?:svg|png)(?:\\?.*)?$"
```
ericwill: I don't think this regex works, I tested it against `file.png` and `file.svg` and it didn't match.

sleshchenko: @vitaliy-guliy quite difficult regexp, could you elaborate about each of the segments?

Like

 ^[-\\._a-zA-Z0-9]+\\.(svg)|(png)$
 
 ^ **->** the beging of the line
 [-\\._a-zA-Z0-9] **->** just file name that could contain -._letters digits
 \\.(svg)|(png) **->** file extension must be .svg or .png
 $ **->** the end

vitaliy-guliy: @akurinnoy helped me to write this regexp, as he mush more experienced.
We have to use double slash when using a regexp as a value in json.
The original regexp is `[-._a-zA-Z0-9]+?\.(?:svg|png)(?:\?.*)?$`. 
Where:
- `[-._a-zA-Z0-9]+?` checks for the file name
- `\.(?:svg|png)` the extension, which can be only .svg or .png
- `(?:\?.*)?` optional query string

sleshchenko: So, query param... It's not really file path but URL?
Then `[-._a-zA-Z0-9]+` does not really make sense... It may work since `.` is not escaped and it includes any character... Or `^` is missing by purpose...
OK, let's discuss offline

vitaliy-guliy: We don't need ^ at the start of the regexp, isn't it?

vitaliy-guliy: You can check the regexp online https://regexr.com/50au9

vitaliy-guliy: > So, query param... It's not really file path but URL?

It looks like it could be a path also. If it's a path, it points on an image on the current plugin registry. So, both variants.

sleshchenko: Oleksii explained to me. 

> I don't think this regex works, I tested it against file.png and file.svg and it didn't match.
@ericwill How did you test? If with some online tool - note that you must replace \\ with \

It works fine for me 
Screenshot_20200313_125417

1. png and svg are allowed
2. .ico or references without file extension are not.
3. Query params are allowed

The only comment that since we miss `^` - user could provide non-valid URL at all, like `$*WUE*#@$&@*#ico.png` and it would be not caught by our check. But it's not issue that is caused by your changes.

sleshchenko: @vitaliy-guliy Please update https://github.com/eclipse/che-plugin-registry#plugin-meta-yaml-structure and add png there as well.

vitaliy-guliy:
@sleshchenko yeah, we have to check the whole URL and path. Let's do it in another PR

vitaliy-guliy:
@ericwill we have to use double slash for regexp in JSON. I didn't know that before adding a regexp with slash.

PR #427, opened 2020-03-12
Validate meta.yaml versions to be strings, not numbers

amisevsk:
### What does this PR do?
Changes the validation performed on meta.yaml version numbers to only accept string values. Accepting number and string as types at the same time can cause subtle issues due to how json parsing works: if a value can be interpreted as a number, it will be; otherwise it is a string. This means: 
- Version `3.2.1` will parse as a string, `3.2` will parse as a number
- Git commit short hashes can *occasionally* be exclusively digits, which results in parsing as a number, e.g. https://github.com/eclipse/che-plugin-registry/commit/946650187395af0a5e13eb91e4b743a3f25a59bb. 

This can trip up implementations that assume they're getting a version string. 

From a design perspective, versions should never be specified as a number, as standard number logic does not apply to versions (1.10 is after 1.9, for example)

### What issue does it address?
The changes in this PR would make the underlying issue in https://github.com/eclipse/che/issues/16340 clear by running `./build/scripts/check_metas_schema.sh`

### DIFF
--- a/build/scripts/meta.yaml.schema
+++ b/build/scripts/meta.yaml.schema
@@ -17,9 +17,8 @@
       "pattern": "^[-._a-zA-Z0-9]+$"
     },
     "version": {
-      "type": ["number", "string"],
-      "pattern": "^[-._a-zA-Z0-9]+$",
-      "minimum": 0
+      "type": ["string"],
+      "pattern": "^[-._a-zA-Z0-9]+$"
     },
     "type": {
       "type": "string",
--- a/build/scripts/util.sh
+++ b/build/scripts/util.sh
@@ -15,8 +15,8 @@
 function evaluate_plugin_id() {
     # yq command to do the same; not used as it is much slower.
     # yq -r '"\(.publisher)/\(.name)/\(.version)"' $1
-    name_field=$(sed -nr 's|^name: ([-.0-9A-Za-z]+)|\1|p' "$1")
-    version_field=$(sed -nr 's|^version: ([-.0-9A-Za-z]+)|\1|p' "$1")
-    publisher_field=$(sed -nr 's|^publisher: ([-.0-9A-Za-z]+)|\1|p' "$1")
+    name_field=$(sed -nr 's|^name: "?([-.0-9A-Za-z]+)"?|\1|p' "$1")
+    version_field=$(sed -nr 's|^version: "?([-.0-9A-Za-z]+)"?|\1|p' "$1")
+    publisher_field=$(sed -nr 's|^publisher: "?([-.0-9A-Za-z]+)"?|\1|p' "$1")
     echo "${publisher_field}/${name_field}/${version_field}"
 }
--- a/v3/plugins/che-incubator/cpptools/0.1/meta.yaml
+++ b/v3/plugins/che-incubator/cpptools/0.1/meta.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 publisher: che-incubator
 name: cpptools
-version: 0.1
+version: "0.1"
 type: VS Code extension
 displayName: C/C++ tools
 title: C/C++ intelliSense, debugging and code browsing


amisevsk:
> Checking plugin 'che-incubator/cpptools/0.1'
>	0.1 is not of type u'string'

Ha, caught one. Fixed.

amisevsk:
Found an issue where the way we get plugin IDs would fail if a field was enclosed in quotation marks; fixed. @ericwill please take another look when you have time.

build/scripts/meta.yaml.schema
```diff
@@ -17,9 +17,8 @@
       "pattern": "^[-._a-zA-Z0-9]+$"
     },
     "version": {
-      "type": ["number", "string"],
-      "pattern": "^[-._a-zA-Z0-9]+$",
-      "minimum": 0
+      "type": ["string"],
+      "pattern": "^[-._a-zA-Z0-9]+$"
```
sleshchenko: Dot means any char, right? So, we need escape it.
Does it make sense to deny -_. in the beginning and in the end?

:warning: The proposal is not tested

```suggestion
      "pattern": "^[a-zA-Z0-9]+([-\\._a-zA-Z0-9][a-zA-Z0-9]+)*$"
```

build/scripts/util.sh
```diff
@@ -15,8 +15,8 @@
 function evaluate_plugin_id() {
     # yq command to do the same; not used as it is much slower.
     # yq -r '"\(.publisher)/\(.name)/\(.version)"' $1
-    name_field=$(sed -nr 's|^name: ([-.0-9A-Za-z]+)|\1|p' "$1")
-    version_field=$(sed -nr 's|^version: ([-.0-9A-Za-z]+)|\1|p' "$1")
-    publisher_field=$(sed -nr 's|^publisher: ([-.0-9A-Za-z]+)|\1|p' "$1")
+    name_field=$(sed -nr 's|^name: "?([-.0-9A-Za-z]+)"?|\1|p' "$1")
+    version_field=$(sed -nr 's|^version: "?([-.0-9A-Za-z]+)"?|\1|p' "$1")
```
sleshchenko: The same comment as for JSON Schema

build/scripts/meta.yaml.schema
amisevsk: Within a character class, `.` has its literal meaning, I believe.

v3/plugins/che-incubator/cpptools/0.1/meta.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 publisher: che-incubator
 name: cpptools
-version: 0.1
+version: "0.1"
```
ericwill: And this one has to be made a string because it's short, i.e. not something like 1.0.9, right?

build/scripts/meta.yaml.schema
sleshchenko: @amisevsk You're right. My comment is fixed now.

And what about ._ at the beginning and the end? Does it make sense?

v3/plugins/che-incubator/cpptools/0.1/meta.yaml
amisevsk: Yup, `0.1` could be parsed as a number, so it is. `1.0.9` cannot, so it isn't and defaults to string. Other fun cases
```yaml
version: 1.0    # => version: 1
version: 0xa    # => version: 10
version: 0b1001 # => version: 9
version: 0777   # => version: 511
```

Considering forcing all strings to be quoted, honestly.

build/scripts/meta.yaml.schema
amisevsk: Yes sure, I'll add it.

v3/plugins/che-incubator/cpptools/0.1/meta.yaml
ericwill: +1 to that

amisevsk:
Updated the regex for version to force versions to start and end with an alphanumeric character.

I think forcing all of name/publisher/version to be quoted should maybe be done in a separate PR; even a straight format of the entire repo is a +1k/-1k line diff.

PR #428, opened 2020-03-20
Added support for s390x

aditijadhav38:
### What does this PR do?
Add support for s390x

### DIFF
--- a/build/dockerfiles/rhel.install.sh
+++ b/build/dockerfiles/rhel.install.sh
@@ -7,8 +7,8 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #
-
-microdnf install -y findutils bash wget yum gzip tar jq python3-six python3-pip skopeo && microdnf -y clean all && \
+microdnf --disablerepo=fedora30-updates --disablerepo=fedora30-secondary-updates install -y findutils bash wget yum gzip tar python3-six python3-pip && microdnf -y clean all && \
+microdnf --enablerepo=fedora30-updates --enablerepo=fedora30-secondary-updates install -y skopeo jq && microdnf update -y skopeo containers-common jq oniguruma && microdnf -y clean all; \
 # install yq (depends on jq and pyyaml - if jq and pyyaml not already installed, this will try to compile it)
 if [[ -f /tmp/root-local.tgz ]] || [[ ${BOOTSTRAP} == "true" ]]; then \
     mkdir -p /root/.local; \
--- a/build/dockerfiles/fedora.repo
+++ b/build/dockerfiles/fedora.repo
@@ -1,0 +1,13 @@
+[fedora30-updates]
+name=Fedora 30 updates - $basearch
+baseurl=http://download.fedoraproject.org/pub/fedora/linux/updates/30/Everything/$basearch/
+gpgcheck=0
+enabled=1
+skip_if_unavailable=True
+
+[fedora30-secondary-updates]
+name=Fedora 30 secondary updates - $basearch
+baseurl=https://dl.fedoraproject.org/pub/fedora-secondary/releases/30/Everything/$basearch/os/
+gpgcheck=0
+enabled=1
+skip_if_unavailable=True
--- a/build/dockerfiles/rhel.Dockerfile
+++ b/build/dockerfiles/rhel.Dockerfile
@@ -43,7 +43,7 @@ ENV BOOTSTRAP=${BOOTSTRAP} \
 # NOTE: uncomment for local build. Must also set full registry path in FROM to registry.redhat.io or registry.access.redhat.com
 # enable rhel 7 or 8 content sets (from Brew) to resolve jq as rpm
 COPY ./build/dockerfiles/content_set*.repo /etc/yum.repos.d/
-
+COPY ./build/dockerfiles/fedora.repo /etc/yum.repos.d/
 COPY ./build/dockerfiles/rhel.install.sh /tmp
 RUN /tmp/rhel.install.sh && rm -f /tmp/rhel.install.sh
 


nickboldt:
Hey, what about also adding support for ppc64le ? 

I mean if you want linux on Z you'll probably want Power too, right?

aditijadhav38:
@nickboldt We are working on adding support for Z only.

build.sh
```diff
@@ -16,7 +16,16 @@ TAG="nightly"
 LATEST_ONLY=false
 USE_DIGESTS=false
 OFFLINE=false
-DOCKERFILE="./build/dockerfiles/Dockerfile"
+architecture=""
+case $(uname -m) in
+       x86_64) architecture="amd64" ;;
+       s390x) architecture="s390x";;
+esac
+if [ $architecture == "amd64" ]; then \
```
benoitf, line 24: hello, I don't see why architecture should be related to the dockerfile that should be used.

Architecture is mostly a parameter/`ARG` of a Dockerfile

build/dockerfiles/rhel.install.sh
```diff
@@ -8,7 +8,36 @@
...
+       microdnf install -y findutils bash wget yum gzip tar jq python3-six python3-pip skopeo && microdnf -y clean all; \
+elif [ $architecture == "s390x" ]; then \
+       microdnf install -y git findutils bash wget yum gzip tar python3-six python3-pip && microdnf -y clean all && \
+       microdnf install -y gcc gcc-c++ make automake; \
+       wget https://github.com/stedolan/jq/releases/download/jq-1.5/jq-1.5.tar.gz; \
+       tar xvzf jq-1.5.tar.gz; \
+       cd jq-1.5 || exit; \
+       ./configure && make && make install; \
+       cd "$HOME" || exit; \
+       export GOVERSION=1.13; \
```
nickboldt, line 26: GO 1.13 isn't supported yet. https://access.redhat.com/containers/#/registry.access.redhat.com/rhel8/go-toolset 
Please use GO 1.12 as that's the latest supported version in RHCC.

build/dockerfiles/rhel.install.sh
```diff
@@ -8,7 +8,36 @@
...
+esac
+if [ $architecture == "amd64" ]; then \
+       microdnf install -y findutils bash wget yum gzip tar jq python3-six python3-pip skopeo && microdnf -y clean all; \
+elif [ $architecture == "s390x" ]; then \
+       microdnf install -y git findutils bash wget yum gzip tar python3-six python3-pip && microdnf -y clean all && \
+       microdnf install -y gcc gcc-c++ make automake; \
+       wget https://github.com/stedolan/jq/releases/download/jq-1.5/jq-1.5.tar.gz; \
+       tar xvzf jq-1.5.tar.gz; \
+       cd jq-1.5 || exit; \
+       ./configure && make && make install; \
```
nickboldt, line 24: Rather than building from source could we use Fedora version?

https://rpmfind.net/linux/rpm2html/search.php?query=jq&submit=Search+...&system=&arch=s390x

build/dockerfiles/rhel.install.sh
```diff
@@ -8,7 +8,36 @@
...
+       tar -xzf go$GOVERSION.linux-s390x.tar.gz; \
+       export GOROOT=$HOME/go; \
+       cp "$HOME"/go/bin/go /usr/bin/; \
+       cp "$HOME"/go/bin/gofmt /usr/bin/; \
+       export CC=gcc; \
+       microdnf install -y gpgme-devel libassuan-devel device-mapper-devel; \
+       export GOPATH=$HOME; \
+       mkdir -p "$GOPATH"/src/github.com/containers; \
+       cd "$GOPATH"/src/github.com/containers || exit; \
+       git clone https://github.com/containers/skopeo; \
```
nickboldt, line 37: rather than build from source, could we use Fedora binary?

https://rpmfind.net/linux/rpm2html/search.php?query=skopeo&submit=Search+...&system=&arch=s390x

build.sh
nickboldt: Well, Dockerfile == alpine, and rhel.Dockerfile == RHEL8. So to build on IBM platforms we need RH OS.

nickboldt: or else you need to have TWO build configurations (two dockerfiles) just to support https://wiki.alpinelinux.org/wiki/S390x/Docker and RHEL8/s390x.

aditijadhav38: @benoitf Because the base images which are being used in Dockerfile does not support s390x whereas the ones used in rhel.Dockerfile does. Hence added architecture related code lines.

build/dockerfiles/rhel.install.sh
aditijadhav38: Thanks for the suggestions. I will work on incorporating these changes.

nickboldt: note that to use this in downstream Brew builds, we would need to prefetch the rpms and stash them in the dist-git repo's lookaside cache, if they don't exist in RHEL content set repos. 

A bit awkward but not impossible.

build.sh
benoitf: it seems to be hackish

because I could use alpine as well with `https://hub.docker.com/r/s390x/alpine/` so architecture does not imply the system.

aditijadhav38: @benoitf alpine is available for s390x but ```registry.centos.org/centos/httpd-24-centos7``` is not available for s390x.
We tried building che-plugin-registry using the Dockerfile but it failed for the above mentioned image.

nickboldt: instead of centos, why not use RHEL?

https://access.redhat.com/containers/?tab=tags&architecture=S390X#/registry.access.redhat.com/rhscl/httpd-24-rhel7

or even better...

https://access.redhat.com/containers/?tab=tags&architecture=S390X#/registry.access.redhat.com/rhel8/httpd-24

(since we use that one in CRW 2.1 downstream)

aditijadhav38:
@benoitf Thanks for updating Dockerfile with alpine/httpd image which reduced the unnecessary changes needed, in order to make it multi-arch.

benoitf:
Hello, it seems PR is not rebased ?

aditijadhav38:
> Hello, it seems PR is not rebased ?

Yes its rebased, please check.

build/dockerfiles/rhel.install.sh
```diff
@@ -8,7 +8,20 @@
...
+architecture=""
+case $(uname -m) in
+       x86_64) architecture="amd64" ;;
+       s390x) architecture="s390x";;
+esac
+if [ $architecture == "amd64" ]; then \
+       microdnf install -y findutils bash wget yum gzip tar jq python3-six python3-pip skopeo && microdnf -y clean all; \
+elif [ $architecture == "s390x" ]; then \
+       microdnf install -y git findutils bash wget yum gzip tar python3-six python3-pip && microdnf -y clean all && \
+       for i in g/glibc-common-2.29-9 g/glibc-all-langpacks-2.29-9 g/glibc-2.29-9 o/oniguruma-6.9.1-2 j/jq-1.6-2 s/skopeo-0.1.36-6.dev.gitd93a581 c/containers-common-0.1.36-6.dev.gitd93a581; do wget https://rpmfind.net/linux/fedora-secondary/releases/30/Everything/s390x/os/Packages/$i.fc30.s390x.rpm; done; \
```
nickboldt, line 20: hardcoding the versions here is super brittle and means you'll have to manually update to newer versions to fix CVE issues. Can you fetch rpms w/o needing to use specific versions?

nickboldt: also it's recommended that you use a version of skopeo >= 0.1.40 so that you have access to the new `skopeo copy --all` command that allows copying multiple images/arches.

You might not actually NEED to do that here, but might as well use a current/recent version so containers and tools are on the same playing field.

aditijadhav38: @nickboldt We did further investigation on your review comments and found that Fedora rpm's till version 30 only can be installed on ubi8 distribution. The later releases of Fedora have zstd compression support in their rpms which ubi8 do not support currently. Thus we cannot install latest Fedora rpms in ubi8 container. Ref. https://bugzilla.redhat.com/show_bug.cgi?id=1715799
Kindly let us know if you have any workaround in your knowledge, else we will have to fallback to 'install from source' strategy.

nickboldt: That's been my experience too... up to and including F30, GOOD. F31+, can't be installed on RHEL8 (at least not on 8.1). 8.2 might add Zstd compression (I don't know, maybe).

Here's how I install Skopeo 0.1.40 from F30 in RHEL7:

```
   yum -y install podman-docker skopeo

    # CRW-563 install skopeo 0.1.40 from Fedora 30 repo
    # NOTE: DO NOT USE Fedora 31 -uses Zstd compression not avail in RHEL 7.7
    # >>> yum error: rpmlib(PayloadIsZstd) <= 5.4.18-1 is needed by skopeo-1:0.1.40-4.fc31.x86_64

    yum-config-manager --add-repo http://download.fedoraproject.org/pub/fedora/linux/updates/30/Everything/x86_64 || true
    yum update -y -q skopeo containers-common --nogpgcheck  --disablerepo="*" \
        --enablerepo=download.fedoraproject.org_pub_fedora_linux_updates_30_Everything_x86_64
    rm -fr /etc/yum.repos.d/download.fedoraproject.org_pub_fedora_linux_*
    skopeo --version
```

benoitf: is there a way to have same instructions between all arch ?

nickboldt: we could enable Fedora 30 repo for `$MY_ARCH` in both cases, yes.

```
   MY_ARCH="$(uname -m)" &&
   if [[ ${MY_ARCH} == "x86_64" ]]; then export MY_ARCH="amd64"; fi && \

    yum-config-manager --add-repo http://download.fedoraproject.org/pub/fedora/linux/updates/30/Everything/${MY_ARCH} || true
    yum update -y -q skopeo containers-common --nogpgcheck  --disablerepo="*" \
        --enablerepo=download.fedoraproject.org_pub_fedora_linux_updates_30_Everything_${MY_ARCH}
```

nickboldt: in the above example we would also install/update `yq`, etc.

aditijadhav38: @nickboldt  We have tried above code given by you, but have observed few things as follows:
1. The repository that you mentioned in the code 'http://download.fedoraproject.org/pub/fedora/linux/updates/30/Everything/ does not have directory for s390x hence facing 404 error.
2. Also faced yum-config-manager:command not found error, hence tried installing it using yum-utils but in ubi8 yum command is not available in ubi-minimal containers.

Hence We use the following approach as workaround:
```echo '[fedora]
name=Fedora $releasever - $basearch
baseurl=https://dl.fedoraproject.org/pub/fedora-secondary/releases/30/Everything/s390x/os/
enabled=1
gpgcheck=0' > /etc/yum.repos.d/fedora.repo
       microdnf install -y glibc-common  glibc-all-langpacks glibc oniguruma containers-common skopeo jq && microdnf update -y skopeo containers-common jq oniguruma
```
The above code is working and has following major changes:
1. Changed repo to 'https://dl.fedoraproject.org/pub/fedora-secondary/releases/30/Everything/s390x/os/' which has all the dependencies and is stable too.

Can we use the above approach for all arch?
 WDYT?

nickboldt: Yes, I don't see why you can't use https://dl.fedoraproject.org/pub/fedora-secondary/releases/30/Everything/ for s390x / ppc64le and http://download.fedoraproject.org/pub/fedora/linux/updates/30/Everything/ for x86_64

build/dockerfiles/rhel.install.sh
```diff
@@ -8,7 +8,25 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
-microdnf install -y findutils bash wget yum gzip tar jq python3-six python3-pip skopeo && microdnf -y clean all && \
+architecture=""
+case $(uname -m) in
+       x86_64) architecture="amd64" ;;
+       s390x) architecture="s390x";;
+esac
+if [ $architecture == "amd64" ]; then \
```
benoitf, line 16: shouldn't the line be the same for all archs ? (no if ?)

aditijadhav38: @benoitf If I use the above line then it throws error for s390x for skopeo and jq before enabling the mentioned repo and if I put the same line after enabling the repo then also it throws following error:
```
error: No available modular metadata for modular package 'python36-3.6.8-2.module+el8.0.0+2975+e0f02136.s390x'; cannot be installed on the system
The command '/bin/sh -c /tmp/rhel.install.sh && rm -f /tmp/rhel.install.sh' returned a non-zero code: 1
```
After investigation found that its because rpm version mismatch between Fedora repo and ubi8 repo.
Hence I kept it separate.

benoitf: I mean, to install skopeo/jq later as well even with amd64

aditijadhav38: Following code works for both amd64 and s390x as well
```
case $(uname -m) in
       x86_64) baseurl="http://download.fedoraproject.org/pub/fedora/linux/updates/30/Everything/x86_64/";;
       s390x) baseurl="https://dl.fedoraproject.org/pub/fedora-secondary/releases/30/Everything/s390x/os/";;
esac

microdnf install -y findutils bash wget yum gzip tar python3-six python3-pip && microdnf -y clean all && \

echo "[fedora]
name=Fedora $releasever - $basearch
baseurl=$baseurl
enabled=1
gpgcheck=0" > /etc/yum.repos.d/fedora.repo

microdnf install -y skopeo jq && microdnf update -y skopeo containers-common jq oniguruma && microdnf -y clean all; \
```
Is this the code change you mean?

nickboldt: LGTM -- both arches are as close as possible to being the same code.

You could simplify further (instead of the case/esac) and just do something like this:

```
arch="$(uname -m)" && baseurl="http://download.fedoraproject.org/pub/fedora/linux/updates/30/Everything/${arch}/" && if [[ ${arch} != "x86_64" ]]; then baseurl="https://dl.fedoraproject.org/pub/fedora-secondary/releases/30/Everything/${arch}/os/"; fi
```

Which is one line and therefore a bit more embeddable in a Dockerfile, if needed

aditijadhav38: @benoitf @nickboldt Updated PR with the code changes. Please review.

nickboldt:
Looks like https://github.com/eclipse/che-devfile-registry/pull/214#pullrequestreview-408457827 is ready to merge... can you prefer the equivalent PR for this repo using that approach?

aditijadhav38:
@nickboldt @benoitf Updated PR with same approach as that of [eclipse/che-devfile-registry#214](https://github.com/eclipse/che-devfile-registry/pull/214). Please review code changes.

PR #429, opened 2020-03-20
Add Git UI Tools plugin

mmorhun:
Signed-off-by: Mykola Morhun <mmorhun@redhat.com>

### What does this PR do?
Adds Git UI Tools Che Theia plugin, which allows using graphical desktop tools in Che workspace.

Some screenshots:

git-ui-tools-screen-1

git-ui-tools-screen-2

git-ui-tools-screen-3


vitaliy-guliy:
It's your custom plugin. I'm not sure we need to add it to the plugin registry.

mmorhun:
Yes it is, but isn't plugin registry for plugins?

v3/plugins/mm4eche/git-ui-tools/1.0.0/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+      public: true
+      targetPort: 6080
+      attributes:
+        protocol: http
+        type: desktop
+        secure: true
+        cookiesAuthEnabled: true
+        discoverable: false
+  extensions:
+    - https://github.com/mmorhun/che-theia-plugin-git-ui-tools/releases/download/0.0.1-pre/che_theia_plugin_git_ui_tools.theia
```
vinokurig: I think it would be better to create a new repository in the https://github.com/che-incubator and keep it there

mmorhun: I think we shouldn't force community to keep their plugins in che incubator.

mmorhun:
> I'd suggest to add this plugin pack in your own devfiles if you want to use it.

Well, I may use it even without a plugin in sidecar. And I am quite familiar with meta.yaml hosted on GitHub. But one of the main goals of any plugin is to be shared with others. How can we reach that goal using personal gists? How can community share their plugins?

I understand your concern about maintaining it. But to my mind Plugins team don't have to do that. The good example is VSCode marketplace. Does VSCode developers support all the plugins there?
In my opinion we should support only essential plugins, mark them as core ones and let others to contribute. It doesn't cost much to host meta.yaml in the registry, especially when each instance of Che has own instance of it.

@ericwill @l0rd @benoitf @gorkem WDTH ?

ericwill:
> > I'd suggest to add this plugin pack in your own devfiles if you want to use it.
> 
> Well, I may use it even without a plugin in sidecar. And I am quite familiar with meta.yaml hosted on GitHub. But one of the main goals of any plugin is to be shared with others. How can we reach that goal using personal gists? How can community share their plugins?

IMO this is what the OpenVSX marketplace is for.

> I understand your concern about maintaining it. But to my mind Plugins team don't have to do that. The good example is VSCode marketplace. Does VSCode developers support all the plugins there?
> In my opinion we should support only essential plugins, mark them as core ones and let others to contribute. It doesn't cost much to host meta.yaml in the registry, especially when each instance of Che has own instance of it.
> 
> @ericwill @l0rd @benoitf @gorkem WDTH ?

VS Code has a marketplace, we do not. Until the OpenVSX marketplace is available I would prefer if we didn't put any "plugin pack" contributions in the registry, except for the official extension packs from VS Code. There is still the issue of overlap, as the Git extension pack will have some of the plugins in this contribution.

PR #442, opened 2020-04-03
Updating model description according to adding lifecycle hooks

vparfonov:
### What does this PR do?

Updating model description according to adding lifecycle hooks https://github.com/eclipse/che-plugin-broker/pull/98


README.md
```diff
@@ -137,6 +137,13 @@ spec:                  # spec (used to be che-plugin.yaml)
...
+      preStop: # the preStop event immediately before the Container is terminated -- see https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/
+        exec: # Executes a specific command, resources consumed by the command are counted against the Container
+          command: ["/bin/sh","-c","/bin/pre-stop.sh"]  #Command is the command line to execute inside the container, the working directory for the command is root ('/') in the container's filesystem. The command is simply exec'd, it is not run inside a shell, so traditional shell instructions ('|', etc) won't work. To use a shell, you need to explicitly call out to that shell. Exit status of 0 is treated as live/healthy and non-zero is unhealthy -- see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#execaction-v1-core
```
amisevsk, line 146: Consider hard-wrapping the descriptions, since by default this results in very long lines (I know we have this problem in other fields as well). Also, I think it's more readable if the commands start in the same column:
```
          command: ["/bin/sh", "-c", "/bin/post-start.sh"]    # Command is the command line to execute inside the container, the working 
                                                              # directory for the command is root ('/') in the container's filesystem. 
                                                              # The command is simply exec'd, it is not run inside a shell, so traditional 
                                                              # shell instructions ('|', etc) won't work. To use a shell, you need to 
                                                              # explicitly call out to that shell. Exit status of 0 is treated as 
                                                              # live/healthy and non-zero is unhealthy -- 
                                                              # see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#execaction-v1-core
      preStop:                                                # the preStop event immediately before the Container is terminated -- see 
                                                              # https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/
        exec:                                                 # Executes a specific command, resources consumed by the command are counted against the Container
          command: ["/bin/sh","-c","/bin/pre-stop.sh"]        # Command is the command line to execute inside the container, the working 
                                                              # directory for the command is root ('/') in the container's filesystem. The command 
                                                              # is simply exec'd, it is not run inside a shell, so traditional shell instructions 
                                                              # ('|', etc) won't work. To use a shell, you need to explicitly call out to that shell. 
                                                              # Exit status of 0 is treated as live/healthy and non-zero is unhealthy -- see 
                                                              # https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#execaction-v1-core
```
Probably we should rewrite this whole section as asciidoc and add better notes, but that's not for this PR.

build/scripts/meta.yaml.schema
```diff
@@ -219,9 +219,38 @@
...
-                      "type": "boolean"
+                    }
+                  }
+                }
+              },
+              "mountSources": {
+                "type": "boolean"
+              }
+            },
+            "lifecycle": {
```
amisevsk, line 230: I'm not sure `lifecycle` is correctly placed here -- shouldn't it be in `.containers.items[].properties`?

sleshchenko: AFAIR it's just the same as K8s container is... And properties - is just map[string,string], it would mean that we would have two properties `postStartCommand`, `preStopCommand` and then we'd transform it into k8s container...
I don't have a strong preference here.

@l0rd WDYT? Is it a good idea to make plugin model closer to k8s?

README.md
vparfonov: fixed

build/scripts/meta.yaml.schema
l0rd: @sleshchenko that's ok for this PR. We need to make sure version 2.0 of the devfile is compatible with it. In particular in version 2.0 we want to define plugins as usual devfile components (of type container/kubernetes/openshift) but defined in an external devfile. And we want that introduce a new devfile section called `events`. cc @davidfestal

ibuziuk:
@ericwill PR check CI is dead broken, would be nice to have it fixed ;-)

ericwill:
> @ericwill PR check CI is dead broken, would be nice to have it fixed ;-)

I believe @JPinkney is still working on them

PR #443, opened 2020-04-08
fix: corrects article in desc of code-server

bartoszmajsak:
### What does this PR do?

Corrects an article in the description of code-server plugin

### DIFF
--- a/v3/plugins/cdr/code-server/next/meta.yaml
+++ b/v3/plugins/cdr/code-server/next/meta.yaml
@@ -5,7 +5,7 @@ version: next
 type: Che Editor
 displayName: code-server
 title: Visual Studio Code - Web (code-server)
-description: A open source distribution of Visual Studio Code as a cloud IDE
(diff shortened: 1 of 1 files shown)


ericwill:
Hi - I'd like to merge this but it looks like you need to sign your commit, so the ECA check can happen.

bartoszmajsak:
> Hi - I'd like to merge this but it looks like you need to sign your commit, so the ECA check can happen.

I made the change through GH and thus it was signed using GH key as you can see on the screenshot:

image

Is it not enough?

bartoszmajsak:
Ah ok, it requires singed-off footer. I will amend then.

ericwill:
You have to close this PR and open another one.

benoitf:
AFAIK It's ok to force push commit without opening a new PR

ericwill:
> AFAIK It's ok to force push commit without opening a new PR

Or that. :)

bartoszmajsak:
@ericwill done :)

PR #446, opened 2020-04-15
feat(alpine): Change parent image to be alpine based as well

benoitf:
### What does this PR do?

It replaces centos/httpd with alpine/httpd allowing to be smaller/updated and also multi-arch

Change-Id: I7a3050fcfdea30fe08e4127cf581713a9cc91b85

### DIFF
--- a/build/dockerfiles/Dockerfile
+++ b/build/dockerfiles/Dockerfile
@@ -11,7 +11,7 @@
 #
 
 # Builder: check meta.yamls and create index.json
-FROM alpine:3.10 AS builder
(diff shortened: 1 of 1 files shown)


rhopp:
@ericwill If you tried to start Che with pluginregistry built from this PR I believe that's fine. If you want me to double check, it would have to wait for tomorrow morning.

ericwill:
I did not check with Che, only that the container is serving the metadata correctly via web browser. It can wait until tomorrow, whenever you have a moment would be great.

PR #450, opened 2020-04-23
Update references to docker images

vitaliy-guliy:

### What does this PR do?

Changes tags on digests when referencing docker images

Only `latest` and `next` versions of the plugins were updated.

The sheet with a list of digests for each tag
https://docs.google.com/spreadsheets/d/191DLLyb01-3M54y0MFElFEGyGrTpAwprJEk6UFNlPdM/edit?usp=sharing

Old references were left in the comments to be able to check.
We can remove them before merging this PR.

Solves the issue https://github.com/eclipse/che/issues/16047

### DIFF
--- a/v3/plugins/broadcommfd/cobol-language-support/0.10.1/meta.yaml
+++ b/v3/plugins/broadcommfd/cobol-language-support/0.10.1/meta.yaml
@@ -11,7 +11,8 @@ repository: https://github.com/eclipse/che-che4z-lsp-for-cobol
 category: Language
 spec:
   containers:
-    - image: quay.io/eclipse/che-sidecar-java:11-f76ca45
+    # - image: quay.io/eclipse/che-sidecar-java:11-f76ca45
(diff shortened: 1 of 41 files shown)


nickboldt:
FWIW,  https://github.com/eclipse/che-plugin-registry/blob/master/build/scripts/write_image_digests.sh is itself out of date compared to the version we're working in for CRW, which adds multiarch support and better fallback behaviour when a listed image is missing (eg., no dotnet on Z, so have to exclude that from devfile and plugin reg, but keep it in for x86_64). https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-plugin-registry/build/scripts/write_image_digests.sh

@amisevsk @ericwill we probably want this pushed upstream for 7.14, but until Che is doing digests, there's no rush (and we're still working out the bugs in CRW 2.2)

ericwill:
The approach has changed from digests to tokens, so let's close this PR until a way forward is discussed/agreed on.

PR #454, opened 2020-05-01
Add new node-debug2 1.33.0a plugin

ericwill:
Revert the previous fix for eclipse/che#16728, and add
a new 1.33.0a plugin to the registry with the correct
node-debug dependency specified.


benoitf:
Build is failing ?

benoitf:
joking this check is always failing :-)

nickboldt:
MAYBE... we should fix that PR build before 7.13.0 is ready to go live in 10 days ... so we have some quality gate? Now that releases happen w/o QE blocking them for 2-4 days, surely having working PR checks is a P1 / SEV1 concern?

ericwill:
> MAYBE... we should fix that PR build before 7.13.0 is ready to go live in 10 days ... so we have some quality gate? Now that releases happen w/o QE blocking them for 2-4 days, surely having working PR checks is a P1 / SEV1 concern?

IIRC Josh is still working on it.

JPinkney:
Yeah, I can't guarantee anything in 10 days. I'm doing that work ontop of my regular work + ontop of maintaining vscode-yaml/yaml-language-server stuff so it usually falls to the side.

With that said,
all PRs have been merged into theia already for it. The only things that actually needs to be done is the workspace needs to be opened and the tests should autorun. Che doesn't clone projects until the workspace is opened in the browser so I'm having to write a python script that will open the project url, login to keycloak and then wait.

Though even when this is all done, tests have only been created for Java. So it would basically only run tests against PR changes to the java plugin.

I'm hoping to at least get it in a working state soon so that I can pass it off for someone to finish

PR #462, opened 2020-05-06
Added support for ppc64le

kumarvikas2020:
Signed-off-by: vikas kumar <kumar.vikas@in.ibm.com>

### What does this PR do?

Added support for ppc64le.

### DIFF
--- a/build/dockerfiles/content_sets_centos8_appstream.repo
+++ b/build/dockerfiles/content_sets_centos8_appstream.repo
@@ -1,5 +1,5 @@
 [AppStream]
 name=CentOS-8 - AppStream
-baseurl=http://mirror.centos.org/centos-8/8/AppStream/x86_64/os/
+baseurl=http://mirror.centos.org/centos-8/8/AppStream/arch/os/
 gpgcheck=0
 enabled=1
--- a/build/dockerfiles/rhel.Dockerfile
+++ b/build/dockerfiles/rhel.Dockerfile
@@ -44,6 +44,8 @@ ENV BOOTSTRAP=${BOOTSTRAP} \
 # enable rhel 7 or 8 content sets (from Brew) to resolve jq as rpm
 COPY ./build/dockerfiles/content_set*.repo /etc/yum.repos.d/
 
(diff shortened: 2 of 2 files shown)


build/dockerfiles/content_sets_centos8_appstream.repo
```diff
@@ -1,5 +1,5 @@
 [AppStream]
 name=CentOS-8 - AppStream
-baseurl=http://mirror.centos.org/centos-8/8/AppStream/x86_64/os/
+baseurl=http://mirror.centos.org/centos-8/8/AppStream/arch/os/
```
nickboldt, line 3: don't you need `$arch` there?

nickboldt:
See also WIP in https://github.com/eclipse/che-plugin-registry/pull/428

build/dockerfiles/content_sets_centos8_appstream.repo
kumarvikas2020: arch is just a placeholder declared inside content_sets_centos8_appstream.repo. Anything inside this file will act as a constant. Using $arch will not act like a normal shell variable.
It needs to be replaced by appropriate architecture (x86_64, ppc64le, etc) depending on the machine on which it is being built.
This replacement will be done by following additional instruction in the rhel.Dockerfile:
``` 
RUN ARCH=`arch` && sed -i "s|arch|${ARCH}|g" /etc/yum.repos.d/content_set*.repo
```

nickboldt: yes but iirc you can embed an OS aware variable and not need to do sed later.

for example...

`baseurl=http://SERVER/content/dist/rhel8/8/$basearch/appstream/os` 

But using `arch` instead of `uname -m` is cool. Didn't know about that command. :D

Anyway please sync your efforts with the s390x team so we're not solving the same problem two ways and ending up with conflicting approaches/PRs. 

cc: @aditijadhav38

aditijadhav38:
@kumarvikas2020 the repository link ```http://mirror.centos.org/centos-8/8/AppStream/``` is not suitable for s390x.
Also could you please check if the solution in #428 works for ppc64le.

kumarvikas2020:
Closing pull request. Will be raising a new one in-line with s390x changes.

PR #464, opened 2020-05-11
Buildkit Dev plugin

vitaliy-guliy:

### What does this PR do?

Adds Buildkit Dev plugin to build docker images inside running container.

Solves issue https://github.com/eclipse/che/issues/16539

Need also to change a pattern for the plugin logo image https://github.com/eclipse/che/issues/16894


vitaliy-guliy:
It looks like we need to change the regexp for the plugin icon verification (remove it at all)
At the moment, it's impossible to use the URI like this `https://avatars3.githubusercontent.com/u/27259197`
```
Checking plugin 'moby/buildkit/0.7.1'
	u'mountPath' is a required property
	u'mountPath' is a required property
	u'https://avatars3.githubusercontent.com/u/27259197' does not match u'[-._a-zA-Z0-9]+?\\.(?:svg|png)(?:\\?.*)?$'
```

vitaliy-guliy:
[ci-test]

ericwill:
Tests are broken for now so they will always fail when triggered.

vitaliy-guliy:
Should we remove the pattern for plugin images or it's better to update it?
```
    "icon": {
      "type": "string",
      "pattern": "[-_a-zA-Z0-9]+?(?:\\.(?:svg|png))?(?:\\?.*)?$"
    },
```

ericwill:
> Should we remove the pattern for plugin images or it's better to update it?
> 
> ```
>     "icon": {
>       "type": "string",
>       "pattern": "[-_a-zA-Z0-9]+?(?:\\.(?:svg|png))?(?:\\?.*)?$"
>     },
> ```
> 
> It looks like it works, but I'm not sure do we need it at all.

We can discuss this change in the issue you filed for it, let's keep it for now in this PR.

rhopp:
@vitaliy-guliy Tests should be fine now AFAIK... You just need to rebase on master.

build/scripts/meta.yaml.schema
```diff
@@ -33,7 +33,7 @@
     },
     "icon": {
       "type": "string",
-      "pattern": "[-._a-zA-Z0-9]+?\\.(?:svg|png)(?:\\?.*)?$"
+      "pattern": "[-_a-zA-Z0-9]+?(?:\\.(?:svg|png))?(?:\\?.*)?$"
```
amisevsk, line 36: I know regexes are generally write-only, but I'm having trouble understanding the change here. If `.svg` or `.png` is optional, it should be removed -- this regex reduces to `[-_a-zA-Z0-9]$`.

vitaliy-guliy: You are right, we can omit checking for the extension here.

vitaliy-guliy:
[ci-test]

vitaliy-guliy:
[ci-test-happy-path]

PR #468, opened 2020-05-20
Update vscode-openshift-connector

ericwill:
Update the latest version to 0.1.5.

Closes eclipse/che#16178


vinokurig:
The plugin is not displayed, however it exists in the plugin's list:
screenshot-che-che 192 168 99 253 nip io-2020 05 21-18_20_20

ericwill:
Do the commands work as expected?

vinokurig:
The commands are not registered as well

ericwill:
Interesting. I just tried testing the Jira/Bitbucket extension in Che and it also was not registered/found. Maybe there is some issue elsewhere?

vinokurig:
I've tried the older version and it worked as usual

ericwill:
Taking another look I see the container is killed off with `OOMKilled` error. I'll investigate.

ericwill:
Upping the memory to 2GB seems to work for now -- the workspace starts and the plugin views/commands are registered.

However as soon as I open the OpenShift connector view, I see an error message:
`Activating extension che-theia.che-openshift-authentication-plugin failed: Cannot read property 'getProviders' of undefined.`

vinokurig:
That's because you are using not the latest editor. Set the editor's version to `latest` to apply the latest plugin API

v3/plugins/redhat/vscode-openshift-connector/0.1.5/meta.yaml
```diff
@@ -0,0 +1,22 @@
...
+description: Interacting with Red Hat OpenShift clusters and providing a streamlined developer experience using Eclipse Che
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/redhat-developer/vscode-openshift-tools
+category: Other
+firstPublicationDate: "2020-05-20"
+spec:
+  containers:
+    - image: "quay.io/eclipse/che-sidecar-openshift-connector:0.1.5-f1b9673"
+      name: "vscode-openshift-connector"
+      memoryLimit: "2G"
```
amisevsk, line 17: `2G` means this version of openshift-connector quadruples the memory requirement from previous; is there not a tighter bound? Maybe `768Mi`?

vinokurig:
`oc` is missing in the sidecar container:
screenshot-che-che 192 168 99 253 nip io-2020 05 22-15_39_20

v3/plugins/redhat/vscode-openshift-connector/0.1.5/meta.yaml
ericwill: It was an experiment to see if the plugin memory limit was an issue (it was), I'll find a tighter bound. It's worth noting though that this version of the plugins uses OpenShift 4, which may have higher memory requirements.

ericwill: Looks like 1500Mi is the sweet spot, 1G or less and the container won't start due to OOM errors.

ericwill:
I tested on hosted Che and all seems to be well, except after I authenticate and return to the workspace I see this error:

`Failed to authenticated the OpenShift connector plugin: error: not a valid URL: parse https://172.30.0.1:443 OpenShift: v3.11.82: net/url: invalid control character in URL
`
@vinokurig since this version of the connector uses OpenShift 4 is it possible some work may need to be done in the authentication plugin?

vinokurig:
The output of the `odo version` command was changed, working on aligning the oauth plugin.

ericwill:
> The output of the `odo version` command was changed, working on aligning the oauth plugin.

Okay, I'll hold off on merging here and will update the PR to include the new auth plugin version once it's released.

vinokurig:
Should be fixed by https://github.com/che-incubator/che-theia-openshift-auth/pull/2, @ericwill could you please review the PR

vinokurig:
It still fails on hosted Che but that's another issue: https://github.com/eclipse/che/issues/17020

PR #471, opened 2020-05-25
Add CPU limitation

vitaliy-guliy:

### What does this PR do?

Adds necessary CPU limitation for the plugin container configuration

Solves https://github.com/eclipse/che/issues/16685

A sheet with a list of images used in plugins
https://docs.google.com/spreadsheets/d/1twoMLRC2A6YjTGiqXGvGvZP_M8GuKGydK75vhXQor2A/edit?usp=sharing

### DIFF
--- a/v3/plugins/eclipse/che-theia/next/meta.yaml
+++ b/v3/plugins/eclipse/che-theia/next/meta.yaml
@@ -58,7 +58,7 @@ spec:
         discoverable: false
   containers:
    - name: theia-ide
-     image: "quay.io/eclipse/che-theia:next"
+     image: quay.io/eclipse/che-theia:next
      env:
          - name: THEIA_PLUGINS
            value: local-dir:///plugins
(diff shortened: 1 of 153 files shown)


l0rd:
A few comments:
1. setting the same limit for every plugin is useless, we already have defaults for this (`CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__LIMIT__CORES`)
2. we should profile a plugins to see how much CPU it uses instead of randomly providing a value: workspaces get broken if we specify wrong values, it's better to set nothing instead.
3. we should specify requests too (in the case of this PR `cpuRequest` but in general `memoryRequest` would be helpful as well)
4. I would recommend to start with che-theia and che-machine-exec only and then, after we get some feedback and we see how it goes, iterate with other plugins as well

v3/plugins/eclipse/che-theia/next/meta.yaml
```diff
@@ -78,10 +78,11 @@ spec:
          - exposedPort: 13131
          - exposedPort: 13132
          - exposedPort: 13133
-     memoryLimit: "512M"
+     memoryLimit: 512M
+     cpuLimit: 500m
```
azatsarynnyy, line 82: Does it really work well?
The benchmarks @vzhukovs did in https://github.com/eclipse/che/issues/18565#issuecomment-753686165 say the opposite - with 500m Che-Theia slows down significantly. See the `Measurement 3`.

ericwill:
Obsoleted by #774

PR #474, opened 2020-05-26
Che4z-2.0.0

filipkroupa:
### What does this PR do?

New versions of plugins:
-  cobol-language-support 0.11.1
-  debugger-for-mainframe 1.1.0
-  hlasm-language-support 0.11.0

Introducing new Che plugin:
 - vscode-extension-for-zowe 1.5.1

Targetting [Milestone 7.14](https://github.com/eclipse/che/milestone/117)


filipkroupa:
I've checked the deprecation example and it seems there is no specific deprecation property - deprecation is determined by presence of deprecate section which contains the autoMigrate properties, is that correct? Also, I was looking for some deprecation guide. I've read [#15693](https://github.com/eclipse/che/issues/15693) but still, I'm not sure what deprecation policy we should consider. We might deprecate all older plugins, or as you suggested, keep 3 versions and deprecate older ones. Also, the autoMigrate puzzles me - what does it do exactly?

ericwill:
> I've checked the deprecation example and it seems there is no specific deprecation property - deprecation is determined by presence of deprecate section which contains the autoMigrate properties, is that correct? 

Correct. You add the deprecate section with the `autoMigrate` and `migrateTo` fields.

>  Also, I was looking for some deprecation guide. I've read [#15693](https://github.com/eclipse/che/issues/15693) but still, I'm not sure what deprecation policy we should consider. We might deprecate all older plugins, or as you suggested, keep 3 versions and deprecate older ones. 

You can choose to deprecate all the older ones if you want, I'm not against that. I am suggesting to deprecate all old versions older than the 3 latest releases we have in the registry. In this case that would only be one version, 0.9.1.


> Also, the autoMigrate puzzles me - what does it do exactly?

It automatically uses the latest (or whatever version is specified in `migrateTo`) plugin if the version selected is deprecated. For example if someone wanted to use vscode-yaml 0.3.0 in Che 7.14, they'd get a message saying it's deprecated. If `autoMigrate` is set to true, then Che would automatically use the latest version, i.e. 0.8.0.

filipkroupa:
@ericwill Thank you for you explanation. I've deprecated the oldest cobol-language-support version per your suggestion and for the future updates, we will be keeping at most 3 non-deprecated plugin versions in the registry.

PR #481, opened 2020-05-28
Fix shellcheck errors in .ci directory

amisevsk:
### What does this PR do?
Resolves the shellcheck errors that have been causing [failed CI jobs](https://app.circleci.com/pipelines/github/eclipse/che-plugin-registry). It'll also stop the email I get on every merged commit :)

There should be no functional changes, but failing this CI step means errors are easier to include in other scripts in the repo.


amisevsk:
@ericwill ping

PR #482, opened 2020-05-28
Updates from CRW 2.2 build scripts

nickboldt:
Updates from CRW 2.2 build scripts (to better align up and downstream repos)

Change-Id: I732f8d891bded2400bac635c43dd3b212e7b13cc


nickboldt:
shellcheck should pass after https://github.com/eclipse/che-plugin-registry/pull/488 (or at least only fail for these files and not unrelated things)

ericwill:
I think with #481 merged this one can go in now too, right @nickboldt ?

build/dockerfiles/content_sets_centos8.repo
```diff
@@ -0,0 +1,11 @@
+[centos-8-for-appstream]
```
benoitf, line 1: as you're moving files, maybe we should move specific files for rhel.Dockerfile to `rhel` folder ?

PR #485, opened 2020-06-02
Extract and use plugin images digest from env.

AndrienkoAleksandr:
### What does this PR do?
https://github.com/eclipse/che/issues/16815

Extract and use plugin images digest from env.

### Depends on:
https://github.com/eclipse/che-operator/pull/274

### DIFF
--- a/build/dockerfiles/entrypoint.sh
+++ b/build/dockerfiles/entrypoint.sh
@@ -48,7 +48,7 @@ IMAGE_REGEX='([[:space:]]*"?)([._:a-zA-Z0-9-]*)/([._a-zA-Z0-9-]*)/([._a-zA-Z0-9-
 # INFO: "=" for base32 it is pad character. If encoded string contains this char(s), then it is always located at the end of the string.
 # Env value it is image with digest to use.
 # Example env variable:
-# RELATED_IMAGE__che_sidecar_clang_plugin_registry_image_HAWTQM3BMRRDGYIK="quay.io/eclipse/che-sidecar-clang@sha256:1c217f34ca69108fdd1ab844c0bcf960edff92519677bde4f8a5f4841b104745"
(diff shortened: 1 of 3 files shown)


build/dockerfiles/entrypoint.sh
```diff
@@ -37,6 +37,47 @@ METAS_DIR="${METAS_DIR:-${DEFAULT_METAS_DIR}}"
 #   \7 - Optional quotation following image reference
 IMAGE_REGEX='([[:space:]]*"?)([._:a-zA-Z0-9-]*)/([._a-zA-Z0-9-]*)/([._a-zA-Z0-9-]*)(@sha256)?:([._a-zA-Z0-9-]*)("?)'
 
+if [[ ! -z $(env | grep .*plugin_registry_image.*) ]];then
```
tolusha, line 40: it is worth adding some comments explaining what the following extract does.

benoitf:
hi, should we use yq to hande yaml instead of grep/sed ?

AndrienkoAleksandr:
> hi, should we use yq to hande yaml instead of grep/sed ?

We, can, but during work on pr https://github.com/eclipse/che-operator/pull/274 I found that yq is much slower, that sed...

AndrienkoAleksandr:
> /bin/bash: .ci/cico-plugin-registry-pr-test.sh: No such file or directory

Ohrimenko1988 Sorry, do we have such test? In the repository I don't see it.

amisevsk:
>> hi, should we use yq to hande yaml instead of grep/sed ?
> We, can, but during work on pr eclipse/che-operator#274 I found that yq is much slower, that sed...

I've found that most of the cost of `yq` is in invoking it, and if you can collapse as much as possible into a single script it's about as fast as a sed/grep solution. That may be infeasible for this task, however.

PR #499, opened 2020-06-15
Contribute a vscode-extensions.json file

ericwill:
This file lists all extensions in the che-plugin-registry. By extension, I mean those Che plugins that use VS Code extensions.

The fields are basic (for now) but are as follows:
* revision: the tag or SHA1 ID of the upstream repository, capturing a certain version of the extension
* vsixDirectory (optional): the sub-directory where the extension is located (some repositories have multiple extensions in multiple folders)
* repository: the location of the upstream repository

This file is needed for eclipse/che#17014, which is part of a greater epic eclipse/che#15819.

### DIFF
--- a/vscode-extensions.json
+++ b/vscode-extensions.json
@@ -1,152 +1,152 @@
 {
   "extensions": [
     {
-      "revision": "6.0.19",
-      "repository": "https://github.com/spgennard/vscode_cobol"
+      "repository": "https://github.com/asciidoctor/asciidoctor-vscode",
+      "revision": "v2.7.7"
     },
     {
-      "revision": "0.11.1",
-      "vsixDirectory": "clients/cobol-lsp-vscode-extension/",
-      "repository": "https://github.com/eclipse/che-che4z-lsp-for-cobol"
+      "repository": "https://github.com/camel-tooling/camel-lsp-client-vscode",
+      "revision": "0.0.24"
     },
     {
-      "revision": "0.9.1",
-      "repository": "https://github.com/eclipse/che-che4z-explorer-for-endevor"
+      "repository": "https://github.com/eclipse/che-che4z-explorer-for-endevor",
+      "revision": "0.9.1"
     },
     {
-      "revision": "0.11.0",
-      "vsixDirectory": "clients/vscode-hlasmplugin/",	
-      "repository": "https://github.com/eclipse/che-che4z-lsp-for-hlasm"
+      "directory": "clients/cobol-lsp-vscode-extension/",
+      "repository": "https://github.com/eclipse/che-che4z-lsp-for-cobol",
+      "revision": "0.11.1"
     },
     {
-      "revision": "2.1650-vsc1.39.2",
-      "repository": "https://github.com/cdr/code-server"
+      "directory": "clients/vscode-hlasmplugin/",
+      "repository": "https://github.com/eclipse/che-che4z-lsp-for-hlasm",
+      "revision": "0.11.0"
     },
     {
-      "revision": "v10.2.1",
-      "repository": "https://github.com/eamodio/vscode-gitlens"
+      "repository": "https://github.com/cdr/code-server",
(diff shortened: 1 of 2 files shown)


vscode-extensions.json
```diff
@@ -0,0 +1,172 @@
...
+      "checkout": "v1.3.0", 
+      "repository": "https://github.com/redhat-developer/vscode-quarkus"
+    },
+    {
+      "version": "0.0.23",
+      "checkout": "master", 
+      "repository": "https://github.com/windup/rhamt-vscode-extension"
+    },
+    {
+      "version": "0.0.22",
```
apupier, line 128: 0.0.24 is now available https://github.com/eclipse/che-plugin-registry/tree/master/v3/plugins/redhat/vscode-apache-camel/0.0.24

vscode-extensions.json
```diff
@@ -0,0 +1,172 @@
...
+      "repository": "https://github.com/camel-tooling/camel-lsp-client-vscode"
+    },
+    {
+      "version": "0.0.13",
```
apupier, line 132: a note to update here or in the other PR https://github.com/eclipse/che-plugin-registry/pull/501 as the version 0.0.14 is on its way

vscode-extensions.json
ericwill: Done!

vscode-extensions.json
ericwill: Done!

l0rd:
Isn't the `version` already available in `package.json` files?
Can we avoid using branches as checkouts (e.g. master) but use only tags or sha1 to allow reproducible builds?

ericwill:
> Isn't the `version` already available in `package.json` files?

Yes, but in this case the version is what we have in the registry, not what is available upstream. We compare this version to what's in the `package.json`.

> Can we avoid using branches as checkouts (e.g. master) but use only tags or sha1 to allow reproducible builds?

Sure, the majority of these use tags to begin with. I will update the branch cases to use SHA1 IDs.

l0rd:
> Yes, but in this case the version is what we have in the registry, not what is available upstream. We compare this version to what's in the `package.json`.

Can we just set the versions in our registry to match the versions of the package.json?

ericwill:
> > Yes, but in this case the version is what we have in the registry, not what is available upstream. We compare this version to what's in the `package.json`.
> 
> Can we just set the versions in our registry to match the versions of the package.json? Having to maintain versions both in the `package.json` and the `vscode-extension.json` looks like an avoidable overhead.

We keep track of the version we have in the registry so that when we upgrade it (someone opens a PR on `vscode-extensions.json`), we can run tests on the new version of the plugin and make sure it works before merging it. We can't just use the upstream version due to API incompatibilities, potential changes to the sidecars, etc. These need to be verified before we can adopt a new version of the plugin in the registry.

ericwill:
@l0rd I've amended the file based on our discussions this morning, please review.

benoitf:
probably `vsixDirectory` could be simplified/just named `directory` as we target VS Code extensions (vsix) in any case

side note: could it be sorted and use repository as first entry ?

Also there is one thing which is not clear to me:
do we retrieve the vsix from the release page or we only build from sources ? 
if we build from source, probably we need to know if npm or yarn needs to be used (or it's assuming based on yarn.lock or package-lock.json)

ericwill:
> probably `vsixDirectory` could be simplified/just named `directory` as we target VS Code extensions (vsix) in any case
> 
> side note: could it be sorted and use repository as first entry ?

Sure I can do that.

> Also there is one thing which is not clear to me:
> do we retrieve the vsix from the release page or we only build from sources ?
> if we build from source, probably we need to know if npm or yarn needs to be used (or it's assuming based on yarn.lock or package-lock.json)

It hasn't been decided yet, so the fields are missing until we land on a way forward. Right now this is the basic version of the file, so we can get the automated report generation working. I can revise this file to add the necessary fields in another PR once we get to the build/publish steps.

PR #503, opened 2020-06-23
Add vscode-tekton to the plugin registry

ericwill:
Initial contribution of vscode-tekton to the plugin registry, based on [che-sidecar-tekton](https://www.github.com/che-dockerfiles/che-sidecar-tekton)

Closes eclipse/che#15752


sunix:
Got that error :(
image

gattytto:
@sunix cancel and it's there anyways 👍

v3/plugins/redhat/vscode-tekton/0.1.0/meta.yaml
```diff
@@ -0,0 +1,21 @@
...
+type: VS Code extension
+displayName: Tekton
+title: Tekton pipelines support
+description: This extension adds rich pipeline support for Tekton
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/redhat-developer/vscode-tekton
+category: Language
+firstPublicationDate: '2020-05-22'
+spec:
+  containers:
```
gattytto, line 14: maybe volume for kube-config so it doesn't have to be re-set after restarting workspace. Same goes for .cache/bazel for bazel extension (this for http_get in skylark)
````yaml
        volumes:
          - mountPath: "/home/theia/.kube"
            name: kube
````
then in cluster workspaces hostpath source
````verilog
ls -alh /data/wksp/workspacesvbcqr4esmvp1h6k/bazelcache                                                                                   total 3.5K
drwxrwxrwx 3 1724 root 3 Jun 23 02:40 .
drwxr-xr-x 8 1724 root 8 Jun 22 23:39 ..
drwxr-xr-x 5 1724 root 5 Jun 23 02:40 _bazel_user
````

````yaml
    volumes:
        - mountPath: "/home/theia/.cache/bazel"
          name: bazelcache
````

File: v3/plugins/redhat/vscode-tekton/0.1.0/meta.yaml [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+    - image: "quay.io/ericwill/che-testing:latest"
+      name: vscode-tekton
+      memoryLimit: '512Mi'
+  extensions:
+    - https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-yaml/vscode-yaml-0.7.2.vsix
+    - https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-kubernetes-tools/vscode-kubernetes-tools-1.1.0.vsix
+    - https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-tekton/vscode-tekton-pipelines-0.1.0.vsix
```
sunix, line 21: 0.2.0 is available https://github.com/redhat-developer/vscode-tekton/releases (btw there is a typo)
Could we use it?

gattytto: oh well 0.0.2 0.2.0 
image

sunix: 0.0.2 seems to be a error but it is 0.2.0 (see tags and download links)

ericwill:
@sunix I build the `0.2.0` version of the plugin using the Jenkins job, so it can be hosted on dl.jboss.org

evidolob:
There is a bug with `tkn` version detection, to make sure that overusing works, you need to use exact version of `tkn`.
It will be fixed with https://github.com/redhat-developer/vscode-tekton/pull/327

Also make sure that you have `kubectl` installed.

ericwill:
@sunix can you please take another look?

v3/plugins/redhat/vscode-tekton/0.2.0/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+repository: https://github.com/redhat-developer/vscode-tekton
+category: Language
+firstPublicationDate: '2020-07-27'
+spec:
+  containers:
+    - image: "quay.io/eclipse/che-sidecar-tekton:0.2.0-7a8b4ad"
+      name: vscode-tekton
+      memoryLimit: '384Mi'
+      volumes:
+        - mountPath: "/home/theia/.kube"
```
sunix, line 19: I am ok with that but, i think all the plugins interacting with kubernetes should use the same volume name. So I suggest https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/vscode-openshift-connector/0.1.5/meta.yaml#L18-L20 to be renamed kube-config instead of oc-config. And adding a volume in the kubernetes plugin with the same name.
BTW any reasons why the kubernetes plugin is having such a high memory request? https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/1.2.0/meta.yaml#L17

ericwill: Why do they all have to have the same name? The OpenShift connector one is for persisting OpenShift configs, this one is for kube configs. IMO the name should be different to reflect that.

The memory requirement is higher for K8s tooling because it requires it, especially when building containers inside a container (dogfooding).

vitaliy-guliy: @sunix sometimes even 1Gb is not enough when building containers from big Dockerfiles

ericwill: @sunix can we merge this one now?

sunix: no ... plz rename the with the same name.
at least we would let the users use the tekton plugin after login with the openshift connector. So i would rename the openshift connector name to match with this one.

sunix: so `kube` or `kube-config` what ever but name the openshift connector volume with `kube` or `kube-config` so if you do login with the openshift connector and connect to one cluster. you would be able to access to tekton plugin ... but if it doesn't work ... i don't see any reason to make that happen. It is our duty to provide to the users, plugins that would live more than 3 months....

ericwill: Done

ericwill:
@sunix I fixed the volume names, please take a look

sunix:
@eric works for me thanks :) :+1:

PR #506, opened 2020-06-24
Add new version of node-debug2 plugin

svor:
### What does this PR do?
Adds new version of node-debug2 plugin;
Fixes version of vscode-node-debug which should be 1.41.1 (https://open-vsx.org/api/ms-vscode/node-debug)


v3/plugins/ms-vscode/node-debug2/1.33.0a/meta.yaml
```diff
@@ -17,5 +17,5 @@ spec:
     name: vscode-node-debug
     memoryLimit: '512Mi'
   extensions:
-  - https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-node-debug/node-debug-1.44.5.vsix
+  - https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-node-debug/node-debug-1.41.1.vsix
```
tsmaeder, line 20: I thought we would do 1.33.0b and automigrate from the previous version to do "bugfix" versions?

tsmaeder: Also, shouldn't we deprecate some old versions when we introduce a new one?

svor: As for me there is no reason to create 1.33.0b as 1.33.0a doesn't work at all. Do you see some benefits to have old 1.33.0a plugin?
As i know we mark some version as deprecate when we have more than 3 version, and node-debug2 plugin has 3 versions.

ericwill: I think it's incorrect to have the 1.33.0a plugin have no extensions from that actual version.In this case I say let's deprecate 1.33.0 entirely (a and b) and have the `autoMigrate` set to `latest`.

svor: Marked 1.33.0a as deprecated and set `automigrate` to latest

PR #507, opened 2020-06-24
Remove CloudShell editor

sleshchenko:
### What does this PR do?
This PR removes CloudShell editor. It's done since there is no any plan where and why it's going to be used.

Even after these changes, it's possible to create CloudShell Workspace with reference to plugin configuration, like:

<details>
<summary>CloudShell Workspace</summary>

```yaml
apiVersion: 1.0.0
metadata:
  name: cloud-shell
components:
  - alias: cloud-shell
    type: cheEditor
    reference: https://raw.githubusercontent.com/eclipse/che-plugin-registry/7.14.3/v3/plugins/eclipse/cloud-shell/nightly/meta.yaml
  - type: dockerimage
    memoryLimit: 256Mi
    alias: dev
    image: 'quay.io/eclipse/che-sidecar-openshift-connector:0.1.2-2601509'
    args: ["tail", "-f", "/dev/null"]
    env:
      - value: '\[\e[34m\]>\[\e[m\]\[\e[33m\]>\[\e[m\]'
        name: PS1
```
</details>

The future of front-end part is supposed to be decided in https://github.com/eclipse/che/issues/17234


benoitf:
cc @l0rd

PR #509, opened 2020-06-25
Update ms-vscode/go plugin

vitaliy-guliy:

### What does this PR do?

- Brings new 0.14.4 version of vscode-go plugin
- Uses newer version of che-sidecar-go container


ericwill:
> Given that the marketplace plugin has changed "publisher"s, should we do the same here?

Good point, the folder name should match the publisher (as should the meta.yaml) -- @vitaliy-guliy can you please fix it. We should probably deprecate the ms-vscode ones and have them point to this new one, too.

vitaliy-guliy:
> > Given that the marketplace plugin has changed "publisher"s, should we do the same here?
> 
> Good point, the folder name should match the publisher (as should the meta.yaml) -- @vitaliy-guliy can you please fix it. We should probably deprecate the ms-vscode ones and have them point to this new one, too.

Yes, I thought about this.

vitaliy-guliy:
> > Good point, the folder name should match the publisher (as should the meta.yaml) -- @vitaliy-guliy can you please fix it. We should probably deprecate the ms-vscode ones and have them point to this new one, too.

Done

PR #518, opened 2020-07-06
Upgrade VS Code Camel language support to 0.0.25 eclipse/che#17340

apupier:

### What does this PR do?

TODO


benoitf:
Hi, could you describe quick action to test LS ?

apupier:
For isntance:
- have a Camel file in the workspace (using the stack for instance https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/apache-camel-springboot/meta.yaml)
- Open the Camel file (for instance https://github.com/jboss-fuse/fuse-rest-http-booster/blob/master/src/main/java/com/redhat/fuse/boosters/rest/http/CamelRouter.java )
- in the parameter of `from` method https://github.com/jboss-fuse/fuse-rest-http-booster/blob/68a6a522e16c08d44e03bfd4d699d2644f3251a0/src/main/java/com/redhat/fuse/boosters/rest/http/CamelRouter.java#L35 there is specific Camel completion.

PR #521, opened 2020-07-07
feat(vscode-extensions): ensure the json file is sorted

benoitf:
### What does this PR do?
Keep the file ordered by github repository name

also attributes are listed in this order: 

- repository
- revision
- directory?

Add a check in the build script to ensure that.

Change-Id: I0b3986f479b5ec401d900f68e3993c1d707669b3


benoitf:
yes I used jq hidden features for that :-)

It's because after two new PRs the file was already unsorted

PR #524, opened 2020-07-07
Fix vscode-extensions.json

ericwill:
Add vscode-didact, fixup golang/vscode-go revision field.


benoitf:
I wanted to add it this is why I sorted the file first

Thx !

PR #527, opened 2020-07-09
Publish report daily at 01:00 UTC

ericwill:
* Added report publish script and GH workflow
* Moved report index.md file to ./report folder


tools/automation/.ci/publish-report.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
benoitf, line 1: Missing copyright header

PR #528, opened 2020-07-10
cobol-language-support 0.12.0

filipkroupa:
### What does this PR do?

Add new versions of plugin cobol-language-support 0.12.0
Deprecate cobol-language-support 0.9.1


benoitf:
Hello @filipkroupa could you bump it there as well https://github.com/eclipse/che-plugin-registry/blob/43411593d20985644334f54d2ee18acb0be4bb23/vscode-extensions.json#L46 ?

thanks

filipkroupa:
@benoitf Sure, I will update this new file, I did not catch you are implementing this change. Does this mean you are moving away from model where metadata for plugin and it's vsix file is provided in PR and you'll be building everything from extension's sources?

benoitf:
@filipkroupa for now it will only generate a report to know if there are new versions of the extension in the registry

after, it should automate the update of the registry

PR #535, opened 2020-07-17
updating didact version to 0.1.16

bfitzpat:
Signed-off-by: bfitzpat@redhat.com <bfitzpat@redhat.com>

### What does this PR do?

Updates the version of the vscode-didact extension from 0.1.15 to 0.1.16, the new version


bfitzpat:
Tested the PR myself on che.openshift.io and looks fine.

apupier:
@ericwill shouldn't we also add the deprecation on the previous version https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/vscode-didact/0.1.15/meta.yaml ? to allow automatic migration to the new version.

sunix:
@apupier IMHO user should use `redhat/vscode-didact/latest` to get the latest version. Deprecation is only if we plan to remove a plugin/version for some reason.

benoitf:
usually previous plugins are deprecated when a new one is added when we do PR against this repo

ericwill:
> usually previous plugins are deprecated when a new one is added when we do PR against this repo

Yes, alternatively I also "allow" 3 releases of a plugin in the registry before I deprecate the oldest ones. Once a plugin is deprecated it's removed after 3 sprints.

For plugins maintained by community members I usually defer to their judgement.

bfitzpat:
> Yes, alternatively I also "allow" 3 releases of a plugin in the registry before I deprecate the oldest ones. Once a plugin is deprecated it's removed after 3 sprints.
> 
> For plugins maintained by community members I usually defer to their judgement.

Now that I know about the policy, I will keep to no more than a couple of releases in the registry at any one time. Thanks for letting me know.

That said, what's the update policy for workspaces that include these plug-ins? Are they automatically updated when a new version becomes available?

bfitzpat:
So perhaps when the `autoMigrate` is working, we can take a spin through our many plug-ins and see what we can deprecate in a smart way as to not break workspaces going forward? @ericwill

PR #539, opened 2020-07-17
Deprecate quarkus-java8 plugins

ericwill:
Java 8 is no longer supported in Quarkus, let's deprecate these plugins and remove them after three sprints.

See eclipse/che#16790


ericwill:
I'm not sure this is the correct way to do it, another option is to remove all quarkus-java8 plugins with a version < 1.4.

ericwill:
Having thought about it more, the 1.2.x and 1.3.0 plugins were deprecated for awhile, so I removed them. For Quarkus Java 8 support users can use 1.3.0a, and use Quarkus on Java 11 for everything else.

I've left the quarkus-java8 1.5.0 plugin deprecated in this PR -- I'll remove it in three sprints.

ericwill:
@sunix please take a look -- let's get this one out of the queue.

v3/plugins/redhat/quarkus-java8/1.5.0/meta.yaml
```diff
@@ -10,6 +10,9 @@ icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
 repository: https://github.com/redhat-developer/vscode-quarkus
 category: Language
 firstPublicationDate: "2020-06-11"
+deprecate:
+  automigrate: true
+  migrateTo: redhat/quarkus-java11/latest
```
nickboldt, line 15: does this also mean updating the quarkus devfile(s) so they're compatible with java11 instead of java8?

ericwill: Yep

PR #549, opened 2020-07-28
Use che-container-tools based sidecar image for vscode-openshift-connector

ericwill:
Adds helm to the sidecar as well. 

Closes eclipse/che#16130


ericwill:
[ci-test]

PR #552, opened 2020-07-30
Default logo for Che plugins

vitaliy-guliy:

### What does this PR do?

Introduces logo for Che plugins.

Instead of having this

Screenshot from 2019-09-10 15-57-47

we can display this (pay attention only on Che logo)

Screenshot from 2019-09-10 15-38-24

There will be another PR (like [this](https://github.com/eclipse/che-plugin-registry/pull/221)) using this image


benoitf:
still wondering if icon should be inside che plugin registry 

it will collide with CodeReady Workspaces downstreaming
will downstream drop the image and add crw icon ?

cc @nickboldt

vitaliy-guliy:
> still wondering if icon should be inside che plugin registry
> 
> it will collide with CodeReady Workspaces downstreaming
> will downstream drop the image and add crw icon ?
> 
> cc @nickboldt

I want Che-Theia to load the image directly from GitHub by URI
https://raw.githubusercontent.com/eclipse/che-plugin-registry/master/v3/images/eclipse-che-logo.png
(will work after merging of this PR)

I don't see any problems with that as we already load such images for GitHub
https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/golang/go/0.14.4/meta.yaml#L9
and from Bitbucket
https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/atlassian/atlascode/2.6.6/meta.yaml#L9

benoitf:
@vitaliy-guliy it's for the branding. go icon is not replaced downstream while che icon will be replaced by codeready icon

so I still don't know if name of the icon needs to be 'che' or 'default-icon' or something that will fit nice for downstream

vitaliy-guliy:
@benoitf ok, understood. Anyway, we can easily rename the image or even upload it somewhere and change all the references.

vitaliy-guliy:
> @vitaliy-guliy it's for the branding. go icon is not replaced downstream while che icon will be replaced by codeready icon
> 
> so I still don't know if name of the icon needs to be 'che' or 'default-icon' or something that will fit nice for downstream

`default-icon.png` will work. We can even load the icon from the plugin registry container, e.g.
```
https://plugin-registry-che.192.168.99.171.nip.io/v3/plugins/default-icon.png
```
Then it should be easy to change the image when building plugin registry container.

But then the next question. How we will refer the image in plugin meta.yaml?
This is one of possible options https://gist.github.com/vitaliy-guliy/f1bb87733b2451b27acccf71256768c6#file-meta-yaml-L10

benoitf:
There is a special handling of prefix 
like
```
relative:extension/images/icon.png
```

vitaliy-guliy:
HTTP looks better for me

ericwill:
> still wondering if icon should be inside che plugin registry
> 
> it will collide with CodeReady Workspaces downstreaming
> will downstream drop the image and add crw icon ?
> 
> cc @nickboldt

AFAIK CRW uses its own icons in its own registry, which is totally separate from ours. It shouldn't cause any more collisions than there are already are.

nickboldt:
> AFAIK CRW uses its own icons in its own registry, which is totally separate from ours. It shouldn't cause any more collisions than there are already are.

We actually disabled the Che icon entirely, replacing it with an empty image:

https://github.com/redhat-developer/codeready-workspaces/commit/54b066ff90e6616f04023ae30bc268d7d7eb5306

nickboldt:
> HTTP url looks better for me

But it's 2020 and http is dead. Long live https/ssl/tls enabled by default!

ericwill:
> > AFAIK CRW uses its own icons in its own registry, which is totally separate from ours. It shouldn't cause any more collisions than there are already are.
> 
> We actually disabled the Che icon entirely, replacing it with an empty image:
> 
> [redhat-developer/codeready-workspaces@54b066f](https://github.com/redhat-developer/codeready-workspaces/commit/54b066ff90e6616f04023ae30bc268d7d7eb5306)
> 
> This was mostly because UX said "if the icons are repetitive and don't provide any useful information, they will simply clutter the UI." (I'm paraphrasing, but that was the gist.)

So this PR doesn't cause any more grief for you? If not, it's a +1 from me.

PR #553, opened 2020-07-31
Add preliminary sidecar fields to vscode-extensions.json

ericwill:
Initial step to add sidecar update checking for the [che-plugin-registry report](https://eclipse.github.io/che-plugin-registry/). Summary of fields:
* **dockerfile**: repository for the Dockerfile that builds the sidecar
* **dockerfileRevision**: (optional) checkout tag/SHA1-ID for the repository to build the sidecar image, in the event the sidecar repo is structured that way

### DIFF
--- a/vscode-extensions.json
+++ b/vscode-extensions.json
@@ -11,7 +11,12 @@
     {
       "repository": "https://github.com/Azure/vscode-kubernetes-tools",
       "revision": "1.2.0",
-      "dockerfile": "https://github.com/che-dockerfiles/che-sidecar-kubernetes-tooling"
+      "sidecar": {
+        "image": "quay.io/eclipse/che-sidecar-kubernetes-tooling:1.2.0-3ea57d5",
+        "source": {
+          "repository": "https://github.com/che-dockerfiles/che-sidecar-kubernetes-tooling"
+        }
+      }
     },
     {
       "repository": "https://github.com/bash-lsp/bash-language-server",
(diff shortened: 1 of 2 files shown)


ericwill:
Also I bumped `vscode-quarkus` because the version was out of date.

ericwill:
The goal, at the outset at least, is to check all sidecars and make sure they are up to date with their repo state. Since almost all sidecars are maintained by us, we can do this by checking the version + SHA1-ID of the sidecar image. When we have more community sidecars we can generalize it to cover more common cases.

vscode-extensions.json
```diff
@@ -19,19 +20,26 @@
     },
     {
       "repository": "https://github.com/bmewburn/vscode-intelephense",
-      "revision": "v1.3.11"
+      "revision": "v1.3.11",
+      "dockerfile": "https://github.com/che-dockerfiles/che-sidecar-php",
```
benoitf, line 24: It's looking strange to have Dockerfile linking a repository no ?

Could we have more structure  like

```yaml
dockerImage :
   - name: quay.io/foo/bar
   - source:
       - url: github.com/foo/bar
       - revision:
       # if not at root /Dockerfile
       - path: ./subfolder/Dockerfile
```

ericwill: We don't need that level of detail right now, since we are maintaining all (but one of) the sidecar images they all have the tag format `version-SHA1 ID`. This allows us to do a simple check, if the `SHA1-ID` from the registry's sidecar doesn't match the `SHA1-ID` from the sidecar repo, then there is a mismatch and it's out of date.

Until we generalize (and enforce) a format for community contributed sidecars, we don't need to actually look at the Dockerfile.

benoitf: Field name is definitely wrong to me. And using same prefix for two vars means that a new level should be added

I would say that it's too fragile to rely on the fact that we maintain all images

ericwill: > Field name is definitely wrong to me. And using same prefix for two vars means that a new level should be added

Sure, a nested JSON object:
```js
{
  "repository": "https://github.com/bmewburn/vscode-intelephense",
  "revision": "v1.3.11",
  "dockerfile":{
    "repository": "https://github.com/che-dockerfiles/che-sidecar-php"
    "revision": "7"
    // optional
    "directory": "./subfolder/foo"
  }
}
```


> I would say that it's too fragile to rely on the fact that we maintain all images

Right now with the exception of one plugin, we maintain all sidecars. And for community contributors I am setting up repos in `che-dockerfiles` where they are using our build/naming system. So I'd say as an interim step we can use this and once we define a proper flow for community contributed sidecars, we can expand/change it. IMO it's better to get at least some automation now, we can always change it later.

PR #558, opened 2020-08-07
Sidecar update checking

ericwill:
This PR adds support for checking of sidecar update status to the nightly che-plugin-registry report.

* New columns for registry sidecar, and upstream sidecar images
* Sidecar images are clickable links, which will take you to the image's repository
* Sidecar images that have a newer upstream version are bolded in the `Upstream Sidecar Image` column
* If a plugin has no sidecar, the sidecar related columns in the report are left blank

### DIFF
--- a/tools/automation/src/check-plugin-updates.ts
+++ b/tools/automation/src/check-plugin-updates.ts
@@ -16,23 +16,37 @@ import * as semver from 'semver';
 
 import simpleGit, { SimpleGit } from 'simple-git';
 
-const ROOT_DIR = '/tmp/repository';
+const EXTENSION_ROOT_DIR = '/tmp/extension_repository';
+const SIDECAR_ROOT_DIR = '/tmp/sidecar_repository';
 
 // struct for the array of vscode-extensions.json file
 export interface Extension {
   repository: string;
   revision: string;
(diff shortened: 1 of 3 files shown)


tools/automation/src/check-plugin-updates.ts
```diff
@@ -95,18 +110,70 @@ export class Report {
...
+              } catch (err) {
+                return this.handleError(entry, `Error executing git rev-parse in sidecar repository: ${err}`);
+              }
+              // Compare versions
+              let upstreamSidecarImageVersion;
+              if (extension.sidecar.source.revision) {
+                upstreamSidecarImageVersion = `${extension.sidecar.source.revision}-${sidecarSHA1}`;
+                entry.sidecarNeedsUpdating = registryImageVersion !== upstreamSidecarImageVersion;
+              } else {
+                const sidecarVersion = (await fs.readFile(path.join(entry.sidecarClonePath, 'VERSION'), 'utf-8')).split(
```
benoitf, line 164: we could maybe use regexp instead of split

ericwill: Agreed, it will be a task for next week

PR #573, opened 2020-08-17
Use node 12 images

ericwill:
Use node 12 as a base image, or in the case of che-sidecar-node, the main image.

### DIFF
--- a/v3/plugins/redhat/php/1.3.11/meta.yaml
+++ b/v3/plugins/redhat/php/1.3.11/meta.yaml
@@ -12,7 +12,7 @@ category: Language
 firstPublicationDate: "2020-05-26"
 spec:
   containers:
-    - image: "quay.io/eclipse/che-sidecar-php:7-fc377b6"
+    - image: "quay.io/eclipse/che-sidecar-php:7-9854327"
       memoryLimit: "1000Mi"
   extensions:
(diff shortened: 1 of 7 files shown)


v3/plugins/redhat/php/1.3.11/meta.yaml
```diff
@@ -12,7 +12,7 @@ category: Language
 firstPublicationDate: "2020-05-26"
 spec:
   containers:
-    - image: "quay.io/eclipse/che-sidecar-php:7-fc377b6"
+    - image: "quay.io/eclipse/che-sidecar-php:7-9854327"
```
benoitf, line 15: do we know why php is requiring nodejs ?

ericwill: I'm not sure TBH, it's been there from the beginning: https://github.com/che-dockerfiles/che-sidecar-php/commit/295ec3296ce7720511e98d5f4736f0ad8028ee98

benoitf: looks like it's a leftover when we had to have nodejs for che-theia endpoint but not related directly to this PR

ericwill: Alright, I'll remove node as the base image and use an alpine image base instead

PR #596, opened 2020-08-27
Add vscode solargraph plugin

liningpan:
### What does this PR do?
Add language support for Ruby using VS Code extension `vscode-solargraph` published by `castwide`

TODO
- Proper memory limit
- Following settings must be configured in `settings.json`, I don't know if there is any way to do this automatically.
    ```
   "solargraph.bundlerPath": "/usr/local/bin/bundle",
   "solargraph.commandPath": "/usr/local/bundle/bin/solargraph",
    ```
- Solargraph language server needs some ruby gems and pre-compile some documentation, do we need post-start script for this.
- GitHub repository of `vscode-solargraph` doesn't seem to be tagged properly.


ericwill:
> TODO
> 
>     * Proper memory limit
> 
>     * Following settings must be configured in `settings.json`, I don't know if there is any way to do this automatically.
>       ```
>       "solargraph.bundlerPath": "/usr/local/bin/bundle",
>       "solargraph.commandPath": "/usr/local/bundle/bin/solargraph",
>       ```

Can you provide more info about these settings and where the `settings.json` file goes?

> 
>     * Solargraph language server needs some ruby gems and pre-compile some documentation, do we need post-start script for this.
> 

I think this can be added in the sidecar.

>     * GitHub repository of `vscode-solargraph` doesn't seem to be tagged properly.

Just use the SHA1 ID of the commit that bumps the version in the upstream `package.json`, in this case it appears to be: https://github.com/castwide/vscode-solargraph/commit/212e76d61ac0790a0f15c4263f1bf54b0f9b80ad

liningpan:
It appears that the default `commandPath` for solargraph executable provided by solargraph plugin does not work properly. So we have to set this parameter manually. I think this is also the cause of the problem described in the original issue [eclipse/che issue #15606](https://github.com/eclipse/che/issues/15606). 

`commandPath` could be configured in user/workspace preference 
solargraph_preference

PS: Boolean parameters are not presented properly.

ericwill:
These sort of preferences can be specified in the `.theia` folder, in a file `settings.json`: https://theia-ide.org/docs/preferences

liningpan:
> 
> 
> These sort of preferences can be specified in the `.theia` folder, in a file `settings.json`: https://theia-ide.org/docs/preferences

Do you mean the `.theia` folder in the workspace? If this is the case, how can we make this as default?

vitaliy-guliy:
@liningpan could you rename the plugin directory to `v3/plugins/castwide/solargraph/0.22.0/meta.yaml ` ?
In the directory name we have to use publisher/name/version exactly as it described in the extension package.json

ericwill:
> > These sort of preferences can be specified in the `.theia` folder, in a file `settings.json`: https://theia-ide.org/docs/preferences
> 
> Do you mean the `.theia` folder in the workspace? If this is the case, how can we make this as default?

You can make it default by specifying it in the devfile you use to start the workspace, see [here](https://github.com/eclipse/che-devfile-registry/blob/3a002814abe67d78da6aa34fd33627628251578c/devfiles/cpp/devfile.yaml#L16) for an example.

liningpan:
I can create a PR for a ruby workspace devfile using this plugin.

ericwill:
@liningpan looks like the ECA check is failing, can you make sure you've signed the CLA, rebase on master, and update this PR?

ericwill (approved):
Thanks for the PR!

PR #597, opened 2020-08-28
Update vscodevim/vim.

monaka:
### What does this PR do?

Adds 2 plugin entries.

* v3/plugins/vscodevim/vim/1.15.0
* v3/plugins/vscodevim/vim/1.16.0


ericwill:
Any reason for the intermediate version (1.15.0)? Why not just jump straight to 1.16.0?

monaka:
@ericwill Because the latest version is not always stable.

monaka:
It looks the language plugin test are broken now. I suppose it can be ignored. I merge this.

PR #603, opened 2020-09-02
Icon/extensions PR check

ericwill:
Contribute a PR check that:
* checks all meta.yaml icon and extensions fields
* raises an error if any icon or vsix file 404's


tools/automation/src/pr-check-icons-extensions-404.ts
```diff
@@ -0,0 +1,94 @@
...
+          metaYamlString = jsyaml.safeLoad(metaYaml, { schema: jsyaml.JSON_SCHEMA });
+        } catch (yamlErr) {
+          extension.error = true;
+          extension.errorMessage = `Error parsing YAML for ${metaYamlFile}: ${yamlErr}`;
+          return extension;
+        }
+
+        if (metaYamlString) {
+          let metaYamlObject: { icon?: string; name: string; version: string; spec?: { extensions?: [string] } };
+          metaYamlObject = JSON.parse(JSON.stringify(metaYamlString, null, 2));
```
benoitf, line 47: curious: why do we stringify the json string ? and why also we need to pretty format it ?

tools/automation/src/pr-check-icons-extensions-404.ts
```diff
@@ -0,0 +1,94 @@
+/********************************************************************************
```
benoitf, line 1: Maybe we need an entry point/ main program with flags
Because it's not natural to share source folders with different main programs.

something more natural is like
```
./tool check-icons
./tool create-report
```

tools/automation/src/pr-check-icons-extensions-404.ts
```diff
@@ -0,0 +1,94 @@
...
+              });
+          }
+
+          // Check the vsix files, if there are any
+          if (metaYamlObject.spec && metaYamlObject.spec.extensions) {
+            extension.vsix = [];
+            for (let vsix of metaYamlObject.spec.extensions) {
+              extension.vsix.push(vsix);
+              await Axios.get(vsix)
+                .then((response) => {})
```
benoitf, line 68: looks odd to use await/then/catch pattern

Usually we get value from await and do something with it

tools/automation/src/pr-check-icons-extensions-404.ts
```diff
@@ -0,0 +1,94 @@
...
+              .then((response) => {})
+              .catch((err) => {
+                extension.error = true;
+                extension.errorMessage = `Failed to download ${metaYamlObject.name}'s icon at ${metaYamlObject.icon}`;
+              });
+          }
+
+          // Check the vsix files, if there are any
+          if (metaYamlObject.spec && metaYamlObject.spec.extensions) {
+            extension.vsix = [];
```
benoitf, line 64: vsix make think it's a single file not an array

benoitf (commented):
few remarks
Nice to have more checks

tools/automation/src/pr-check-icons-extensions-404.ts
ericwill: Maybe this belongs in its own node project under `/.ci/pr-checks/`?

tools/automation/src/pr-check-icons-extensions-404.ts
ericwill: I tried that way, but when the response times our or failed I was getting unhandled promise issues. I'll take another look.

tools/automation/src/pr-check-icons-extensions-404.ts
ericwill: AFAIU, js-yaml returns any of the following [string | object | undefined], so it gets converted to a JSON string then an object we can use.

I can remove the pretty formatting though that was unneeded.

.ci/pr-check/src/pr-check.ts
```diff
@@ -90,5 +91,13 @@ async function checkAllPlugins() {
...
 (async (): Promise<void> => {
-  await checkAllPlugins();
+  let myArgs = process.argv.slice(2);
+  switch (myArgs[0]) {
+    case 'icons-extensions-404':
+      await iconsExtensions404Check();
+      break;
+    case 'all':
+    default:
+      console.log("all")
```
benoitf, line 101: shouldn't 'all' run also extension404 check ?

benoitf (approved):

.ci/pr-check/src/pr-check.ts
ericwill: Oops, leftover testing stuff :)

PR #611, opened 2020-09-09
Add vscode-kotlin extension

blues-man:
Signed-off-by: Natale Vinto <nvinto@redhat.com>

### What does this PR do?
Add language support for Kotlin using VS Code extension `fwcd.kotlin`

TODO
* ~~Resolve issue on extension activation~~


blues-man:
Hello, reopening this with new extension from [fwcd](https://github.com/fwcd/vscode-kotlin/).

I would like to discuss here this issue I found in extension activation giving this error:
image

```Activating extension 'Kotlin' failed: ENOENT: no such file or directory,
mkdir /home/theia/home/theia/fwcd.kotlin```

It looks like it is trying to create the extension directory into Theria $HOME$HOME, but it is not using `mkdir -p`

Is that expected?

File: vscode-extensions.json [hunk 1]
```diff
@@ -367,6 +367,16 @@
           "repository": "https://github.com/che-dockerfiles/che-sidecar-protobuf"
         }
       }
+    },
+    {
+      "repository": "https://github.com/fwcd/vscode-kotlin/",
+      "revision": "v0.2.18",
```
ericwill, line 373: No "v", the upstream repo just used `0.2.18` as the release tag.

gattytto:
@blues-man you have the extension trying to create a folder after being initialized by che:
https://github.com/fwcd/vscode-kotlin/blob/bd965d60fc20a08a32d80f3466c6227dd1b221aa/src/serverDownloader.ts#L61
````typescript
private async downloadServer(downloadUrl: string, version: string, status: Status): Promise<void> {
        if (!(await fsExists(this.installDir))) {
            await fs.promises.mkdir(this.installDir, { recursive: true });
        }
````

the commands (like mkdir) run by the extension are executed within its sidecar context, so if you mkdir -p the /home/theia/home/theia/fwcd.kotlin it results in this being downloaded inside the sidecar:
image
image


Conclusion: try to find the settings in the extensions package.json that will let you customize the path where the extension is trying to download the language server. Or debug the extension to see how you can "help" it figure the right path string (seeing what variables can be used as a base for the folder calculation).

File: v3/plugins/fwcd/vscode-kotlin/0.2.18/meta.yaml [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+displayName: Kotlin
+title: Kotlin tooling for VS Code
+description: This plug-in provides support for Kotlin development.
+icon: https://raw.githubusercontent.com/JetBrains/logos/master/web/kotlin/kotlin.svg
+repository: https://github.com/fwcd/vscode-kotlin/
+category: Language
+spec:
+  containers:
+    - image: "quay.io/eclipse/che-sidecar-java:11-7bd8c8c"
+      name: vscode-quarkus-kotlin
```
ericwill, line 15: If quarkus isn't part of this plugin I'd say just name this `vscode-kotlin`

gattytto: I can confirm there's no result for "quarkus" search within the kotlin extension source repo. So I don't think there's a direct/indirect dependency on it.

gattytto:
if this to be solved serving the language and debug servers inside the sidecar, this is what I did:
1) open the project using the kotlin extension, it will fail
2) open a shell to the kotlin sidecar (now being based on java11), mkdir -p /home/theia/home/theia/fwcd.kotlin
3) go to workspaces menu in che, and come back to the project (so extension reloads). Now it will offer to download stuff
4) once stuff is downloaded to the sidecar it can be copied and referenced using the following settings: (locations are just an example)
````json
{
    "kotlin.languageServer.path": "/home/theia/kotlin_ls/fwcd.kotlin/langServerInstall/server/bin/kotlin-language-server",
    "kotlin.debugAdapter.path": "/home/theia/kotlin_ls/fwcd.kotlin/debugAdapterInstall/adapter/bin/kotlin-debug-adapter"
}
````

image

File: v3/plugins/fwcd/vscode-kotlin/0.2.18/meta.yaml (hunk 2 above)
Comment (blues-man) [reply v3/plugins/fwcd/vscode-kotlin/0.2.18/meta.yaml]: updated thanks, the quarkus part is only for the devfile I pushed also to `che-devfile-registry` using this plugin together with quarkus one

**Comment by blues-man** [comment]
I was able to workaround it adding a volume in the sidecar into plugin's `meta.yaml` like this:

```
spec:
  containers:
    - image: "quay.io/eclipse/che-sidecar-java:11-7bd8c8c"
      name: vscode-kotlin
      memoryLimit: "1500Mi"
      volumes:
      - mountPath: "/home/theia/.m2"
        name: m2
      - mountPath: "/home/theia/home/theia"
        name: vscode-theia-workaround
```

Don't know if this could be a suitable option?

**Comment by gattytto** [comment]
> @gattytto I don't get why `langServerInstallDir` here https://github.com/fwcd/vscode-kotlin/blob/7f2ef9c81bf6109d88a37397d3add54d0a6b1d18/src/languageSetup.ts#L20 is landing to this path `/home/theia/home/theia` containing `/home/theia` twice, is that expected path? That seems the issue

for starters, the runtime execution provider that runs the mkdir command prepends /home/theia by default, then the extension tries to download a zip file to that location, which didn't get created for some reason. 
all this behauvoir is not needed in che design, that part could be just stripped from the extension code and pre-allocate the binaries somewhere in the sidecar filesystem, then from within devfile (or plugin meta-yaml which is not implemented) send that settings to the extension.

**Comment by gattytto** [comment]
@blues-man I'm trying to build an nadroid native bin using kotlin and gradlew and it's asking me for the and sdk, do you think there could be a branch of the cli image with android sdks included? how is the android dependencies to be dealt with when it comes to language-server and include tracking? I'm not good with java stuff and my android is rusty
image

**Comment by gattytto** [comment]
I have now also tried to add the java extension as you did in your devfile-registry pull request, the extension has the same behavior around the home/theia duplication @ericwill this seems to affect both kotlin and java-quarkus extensions
image

**Comment by blues-man** [comment]
The workaround mentioned above worked, however I can't have `javax.ws` resolved, could you have that in your test @gattytto ?

**Comment by blues-man** [comment]
It looks it misses `mvn` from the sidecar:

```
[Info  - 12:08:25 AM] main      Adding script definitions [ScriptTemplateWithArgs]
[Info  - 12:08:25 AM] main      Connected to client
[Info  - 12:08:25 AM] client    Adding workspace file:///projects to source path
[Info  - 12:08:25 AM] client    Adding .../rest/GreetingResource.kt, .../rest/ Greeting.kt, .../rest/GreetingResourceTest.kt, .../rest/NativeGreetingResourceIT.kt under /projects to source path
[Info  - 12:08:25 AM] client    Searching for dependencies and Java sources in workspace root /projects
[Warn  - 12:08:25 AM] client    Could not resolve classpath using Maven: Unable to find the 'mvn' command
[Info  - 12:08:25 AM] client    Could not resolve kotlin-stdlib using Maven: null
[Info  - 12:08:25 AM] client    Could not resolve kotlin-stdlib using Gradle: /home/theia/.gradle/caches
[Info  - 12:08:25 AM] client    Could not resolve kotlin-stdlib using Maven: null
[Info  - 12:08:25 AM] client    Could not resolve kotlin-stdlib using Gradle: /home/theia/.gradle/caches
[Info  - 12:08:25 AM] client    Update build script path
```

**Comment by blues-man** [comment]
I managed to have it working with 2 adds:

- Adding a volume into the sidecar for `/home/theia/home/theia` path
- Adding a custom sidecar with maven extending che-sidecar-java with maven package: [che-sidecar-maven](https://github.com/blues-man/che-sidecar-maven)

image
image

**Comment by gattytto** [comment]
@blues-man not implemented?
image

**Comment by gattytto** [comment]
https://github.com/fwcd/vscode-kotlin/issues/29

**Comment by gattytto** [comment]
@blues-man can you test the imports with a gradle based project using the gradle sidecar?

**Comment by gattytto** [comment]
I have tested the following changes to the plugin.yaml for having a working gradle example:
````yaml
apiVersion: v2
publisher: fwcd
name: vscode-kotlin
version: 0.2.18
type: VS Code extension
displayName: Kotlin
title: Kotlin tooling for VS Code
description: This plug-in provides support for Kotlin development.
icon: https://raw.githubusercontent.com/JetBrains/logos/master/web/kotlin/kotlin.svg
repository: https://github.com/fwcd/vscode-kotlin/
category: Language
spec:
  containers:
    - image: "quay.io/eclipse/che-sidecar-java:11-7bd8c8c"
      name: vscode-quarkus-kotlin
      memoryLimit: "2500Mi"
      env:
      - value: "/home/theia/.gradle"
        name: GRADLE_USER_HOME
      - value: "/home/theia"
        name: HOME
      volumes:
      - mountPath: "/home/theia/.m2"
        name: m2
      - mountPath: "/home/theia/home/theia"
        name: vscode-theia-workaround
  extensions:
    - https://github.com/fwcd/vscode-kotlin/releases/download/0.2.18/kotlin-0.2.18.vsix
````
image

**Comment by blues-man** [comment]
Hello @gattytto thanks for sharing it! I tried generating the Kotlin Quarkus quickstart as grandle project and that works fine with those env you tried, in this way we don't have to use any custom sidecar and any Devfile would work straightforward **if** and only if, they are Gradle project.

I can update the PR code with that, I would like to know what's the best option here?

**Comment by gattytto** [comment]
@blues-man I think the extension is not fully functional, but it's hard for me to reproduce, could you please come to mattermost so we can chat about it more fluidly?.

**Comment by ericwill** [comment]
> @blues-man I think the extension is not fully functional, but it's hard for me to reproduce

Which parts are missing?

File: v3/plugins/fwcd/vscode-kotlin/0.2.18/meta.yaml [hunk 3]
```diff
@@ -0,0 +1,23 @@
...
+version: 0.2.18
+type: VS Code extension
+displayName: Kotlin
+title: Kotlin tooling for VS Code
+description: This plug-in provides support for Kotlin development.
+icon: https://raw.githubusercontent.com/JetBrains/logos/master/web/kotlin/kotlin.svg
+repository: https://github.com/fwcd/vscode-kotlin/
+category: Language
+spec:
+  containers:
```
gattytto, line 13:
````yaml
env:
      - value: "/home/theia/.gradle"
        name: GRADLE_USER_HOME
````

gattytto:
> > @blues-man I think the extension is not fully functional, but it's hard for me to reproduce
> 
> Which parts are missing?

outline, kotiln dsl (.kt files) linting. 

image
image

gattytto:
there's a setting I have to try next 'kotlin.compiler.jvm.target' : "1.8" .. putting it here for the record

gattytto:
so far using the following devfile (and making sure the setting is corretly put in /home/theia/.theia/settings.json)
````yaml
metadata:
  name: kotlin
projects:
  - name: kotlin-examples
    source:
      location: 'https://github.com/Kotlin/kotlin-examples'
      type: git
    clonePath: kotlin
components:
  - id: eclipse/che-theia/next
    type: cheEditor
  - type: chePlugin
    reference: >-
      https://raw.githubusercontent.com/gattytto/cheplugins/master/kotlin/0.2.18/meta.yaml
    preferences:
      kotlin.compiler.jvm.target: "1.8"
    alias: kotlin
apiVersion: 1.0.0
````

produces the following results: Although the command-line part works ok within the sidecar terminal (no need for a CLI), the extension doesn't resolve the dokka import. 
making sure 'kotlin.compiler.jvm.target' : "1.8" is set in the theia settings makes other warnings around the build.gradle.kts files go away. The test also uses a specific sidecar (java11-gradle diversion), with the kotlinc binary added
image

gattytto:
[this](https://github.com/gattytto/che-sidecar-java/blob/e04af199ff703f40649f5beda8ac11bb3fb85ac1/Dockerfile#L45) line helps to prevent the language server from breaking after issuing the F1=>"Kotlin: Restart the Language Server".

gattytto:
clicking an outline item (now works) proposes the fix that actually erases the word:
image
image

blues-man:
I think that this extension, notwithstanding some bugs yet to correct, is the most complete for VSCode, maybe we can ship the che plugin with the maven sidecar and the gradle envs, so it will support both, releasing this version and updating with new ones.

What do you think?

gattytto:
@blues-man I think that the cleaner solution would be to have [this](https://github.com/eclipse/che/issues/17975) implemented first

gattytto:
gladly, time to revive this @ericwill @blues-man

gattytto:
> > > @blues-man I think the extension is not fully functional, but it's hard for me to reproduce
> > 
> > 
> > Which parts are missing?
> 
> outline, kotiln dsl (.kts files) linting.
> 
> image
> image

TODO: check .kts support status

PR #612, opened 2020-09-09
Update build script to use python 3

ericwill:
This fixes the CircleCI failure.


ericwill:
The reason the build was failing earlier: `pyrsistent requires Python '>=3.5' but the running Python is 2.7.18`

benoitf (approved):

PR #614, opened 2020-09-09
Remove deprecated plugins

ericwill:
Some house cleaning -- remove plugins that have been deprecated for more than three sprints.

See eclipse/che#16446


benoitf (approved):

benoitf:
##[error]Process completed with exit code 1.

For icon check ?

benoitf:
should we deprecate `redhat/java` as well (and then migrate them to java8 or java11) ?

ericwill:
> should we deprecate `redhat/java` as well (and then migrate them to java8 or java11) ?

Yes, in a separate PR I think. This one is only for removals.

> ##[error]Process completed with exit code 1.
> For icon check ?

I just rebased the patch, it should have the fix now.

svor (approved):

PR #618, opened 2020-09-11
Update Metals, the Scala language server to version 1.9.3

sideeffffect:
Signed-off-by: sideeffffect <ondra.pelech@gmail.com>


benoitf:
hi, please update https://github.com/eclipse/che-plugin-registry/blob/master/vscode-extensions.json#L296-L301 as well

benoitf:
also your ECA check is failing (eclipse agreement)

sideeffffect:
There must be something wrong with that check
I've signed a long time ago
And it passed just now here, for example https://github.com/eclipse/che-devfile-registry/pull/285

benoitf:
@sideeffffect there is no error on the PR check, you've used different addresses
(commit author and signed-off email need to be the same)

image

benoitf:
also could you deprecate previous versions (in meta.yaml)
like https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/java11/0.50.0/meta.yaml#L13-L15

`scalameta/metals/latest` in your case

sideeffffect:
done, thanks!

benoitf (approved):

PR #620, opened 2020-09-11
Use yarn workspaces and improve icons-extensions-404 PR check

ericwill:
* Use yarn workspaces as we have multiple JS projects now
* Improve the icons/extensions 404 check to use HEAD requests

Fixes eclipse/che#17838


benoitf (approved):
root package.json is usually named with 'monorepo' in it + url + licenses, etc
Also would be nice to bring there the common dependencies to avoid to duplicate it
But it can be done later

.ci/pr-check/src/pr-check.ts
```diff
@@ -64,11 +64,21 @@ async function iconsExtensions404Check() {
             extension.vsixList = [];
             for (let vsix of metaYamlObject.spec.extensions) {
               extension.vsixList.push(vsix);
+              let hostedOnGitHub = vsix.includes("github.com")
               try {
-                await Axios.get(vsix);
+                // GitHub release assets redirect to AWS, but AWS returns 403 for HEAD requests.
```
benoitf, line 69: ok not so nice then aws then !

ericwill: Nope, it's a known issue apparently :/

ericwill:
> root package.json is usually named with 'monorepo' in it + url + licenses, etc

What do you mean by monorepo?

benoitf:
monorepo name like there https://github.com/eclipse-theia/theia/blob/master/package.json#L3

benoitf:
maybe we should move `.ci/pr-check` from `.ci` folder to `tools` folder as well to have something more unified for contributors.

ericwill:
> monorepo name like there https://github.com/eclipse-theia/theia/blob/master/package.json#L3

Will fix it today.


> maybe we should move `.ci/pr-check` from `.ci` folder to `tools` folder as well to have something more unified for contributors.

IMO PR checks are different from the plugin-registry-specific automation that goes in `/tools/automation`. So I would rather keep them separate.

benoitf:
> IMO PR checks are different from the plugin-registry-specific automation that goes in /tools/automation. So I would rather keep them separate.

it's just that it's not obvious to have "source code' in a hidden folder (`.ci`) and we've typescript code, so this is why I suggesting to move it to `tools` but not merge it into automation

like `tools/pr-check` and `tools/automation`

ericwill:
> > IMO PR checks are different from the plugin-registry-specific automation that goes in /tools/automation. So I would rather keep them separate.
> 
> it's just that it's not obvious to have "source code' in a hidden folder (`.ci`) and we've typescript code, so this is why I suggesting to move it to `tools` but not merge it into automation
> 
> like `tools/pr-check` and `tools/automation`

Ah, okay that sounds fair. I'll change it :+1:

PR #637, opened 2020-10-05
Openshift connector 0.2.0

ericwill:
### What does this PR do?


ericwill:
Tested on che.openshift.io and it seems to work without issue

svor (approved):

PR #668, opened 2020-11-05
[WIP] Upgrade VS Code Metals, the Scala language client, to version 1.9.5

sideeffffect:
Signed-off-by: sideeffffect <ondra.pelech@gmail.com>

PR #682, opened 2020-11-10
fix cachedirs ownership

gattytto:


ericwill:
This fixes https://github.com/che-dockerfiles/che-sidecar-scala/issues/10, right?

gattytto:
@ericwill yes partially, then https://github.com/eclipse/che-plugin-registry/pull/668/files goes around updating the extension version, deprecating the old one and removing cache volumes to keep sidecar contents, and then when this one is merged that one should be updated to reflect the new sidecar id

ericwill (approved):

PR #685, opened 2020-11-10
Upgrade VS Code Metals, the Scala language client, to version 1.9.6

sideeffffect:
Signed-off-by: sideeffffect <ondra.pelech@gmail.com>


vscode-extensions.json
```diff
@@ -344,7 +344,7 @@
     },
     {
       "repository": "https://github.com/scalameta/metals-vscode",
-      "revision": "v1.9.4",
+      "revision": "v1.9.6",
       "sidecar": {
         "image": "quay.io/eclipse/che-sidecar-scala:0.9.0-b92a5f9",
```
gattytto, line 349: quay.io/eclipse/che-plugin-sidecar:scala-0275d91

v3/plugins/scalameta/metals/1.9.6/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+displayName: Scala (Metals)
+title: Scala language server with rich IDE features
+description: This plug-in provides support for Scala development via the Metals language server.
+icon: https://raw.githubusercontent.com/scalameta/metals-vscode/v1.9.6/images/logo.png
+repository: https://github.com/scalameta/metals-vscode
+category: Language
+firstPublicationDate: "2020-11-10"
+spec:
+  containers:
+    - image: "quay.io/eclipse/che-sidecar-scala:0.9.0-b92a5f9"
```
gattytto, line 15: quay.io/eclipse/che-plugin-sidecar:scala-0275d91
that's the latest (simplified) one you committed. 

then the volumes below should be gone to keep the ones coming from the sidecar

sideeffffect:
This is ready to be merged.
I think we want to keep the volumes, because we want to share the JARs in `~/.ivy2` (& co.) between the Metals container and the other containers, right?

ericwill (commented):
We are only keeping the latest version of every plugin in the registry these days. Just update the existing `meta.yaml` to the new version, as well as `latest.txt` and update the entry in `vscode-extensions.json`

sideeffffect:
@ericwill good like this?

ericwill (approved):
Exactly like that :+1:

PR #701, opened 2020-11-18
Update MTA plugin to 0.0.56 and migrate to quay image

johnsteele:
Signed-off-by: John Steele <programjsteele@gmail.com>

### PR Checklist

[As the author of this Pull Request I made sure that:](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#pull-request-template-and-its-checklist)

- [ ] [The Eclipse Contributor Agreement is valid](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-eclipse-contributor-agreement-is-valid)
- [ ] [Code produced is complete](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-produced-is-complete)
- [ ] [Code builds without errors](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-builds-without-errors)
- [ ] [Tests are covering the bugfix](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#tests-are-covering-the-bugfix)
- [ ] [The repository devfile is up to date and works](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-repository-devfile-is-up-to-date-and-works)
- [ ] [Sections `What issues does this PR fix or reference` and `How to test this PR` completed](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#sections-what-issues-does-this-pr-fix-or-reference-and-how-to-test-this-pr-completed)
- [ ] [Relevant user documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [ ] [Relevant contributing documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [ ] [CI/CD changes implemented, documented and communicated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#cicd-changes-implemented-documented-and-communicated)

### Reviewers

Reviewers, please comment how you tested the PR when approving it.

PR #702, opened 2020-11-18
18292 - switch to use quay copies of...

nickboldt:
18292 - switch to use quay copies of docker.io images

Change-Id: I3a711926c1c420eeaa4add06265cb3fb4bb294db


nickboldt:
Questions:

a) is it OK to use a `latest` tag?

```
- image: "quay.io/windupeng/mta-vscode-extension:latest"
```

b) Why do we have 3 versions of code-server? Can we trash the older ones and just keep next? Or even better, should we move up to 3.7.1 and start having *released* versions in ther instead of a nightly/next/latest?

ericwill:
> Questions:
> 
> a) is it OK to use a `latest` tag?
> 
> ```
> - image: "quay.io/windupeng/mta-vscode-extension:latest"
> ```

For a community contribution it's okay with me. There is actually a [PR](https://github.com/eclipse/che-plugin-registry/pull/701) open right now to update this plugin, and there may be a PR coming to integrate the sidecar into this repo (meaning no latest tags, `VERSION` + SHA1 tagging system).

> b) Why do we have 3 versions of code-server? Can we trash the older ones and just keep next? Or even better, should we move up to 3.7.1 and start having _released_ versions in ther instead of a nightly/next/latest?

Keep the latest version (`3.6.2`) and the others (including `next`) can be removed.

ericwill (approved):

benoitf (commented):
I'm -1 on these changes
Plugins are coming from contributors so there is no need to fork

Also for some of them we can fix it at publishing source

Please revert, agreement was only for the images bootstraping che

benoitf:
but if you log-in, then pull would work then you don't need that

benoitf:
@nickboldt please revert, remove these editor images from quay.io/ecliose and drop this commit from 7.22.x branch
I asked the revert yesterday

benoitf:
or I'll remove the image myself as https://github.com/eclipse/che-release/pull/9 is not approved

nickboldt:
partially reverted as #706, then https://github.com/eclipse/che-release/pull/9 merged with approval.

PR #716, opened 2020-11-26
Adding 'cpuLimit', `cpuRequest` and `memoryRequest` to the che-theia plugin meta.yaml

ibuziuk:
### What does this PR do?
Adding 'cpuLimit' to the che-theia plugin meta.yaml

theia is not designed to take advantage of multiple cores and I believe the best practice will be to keep CPU at 1 or below

https://github.com/eclipse/che/issues/18472

### PR Checklist

### Reviewers

Verified on:

- [ ] kind
- [ ] minikube
- [ ] crc
- [x] openshift dev-cluster (works when non-milicores values are defined e.g. `1` / `0.03` 
  - see https://github.com/eclipse/che-plugin-registry/pull/716#issuecomment-735767175
- [ ] RHPDS
- [x] che.openshift.io

### DIFF
--- a/v3/plugins/eclipse/che-theia/next/meta.yaml
+++ b/v3/plugins/eclipse/che-theia/next/meta.yaml
@@ -89,6 +89,9 @@ spec:
          - exposedPort: 13132
          - exposedPort: 13133
      memoryLimit: "512M"
+     memoryRequest: "512M"
+     cpuLimit: "1000m"
+     cpuRequest: "30m"
   initContainers:
   - name: remote-runtime-injector
     image: "quay.io/eclipse/che-theia-endpoint-runtime-binary:next"


ericwill:
Some questions:
* What problem are we trying to solve? I read through the initial issue but don't see the exact problem we are trying to fix
* Does this affect things like self-hosting, as well as plugin actions/tasks/commands? There are some plugins that run in the theia container, so looking at theia startup time isn't the only factor.

l0rd:
@ericwill the problem is that if we do not specify a CPU limit in Theia meta.yaml, a default value will be set. On dev clusters (minikube, openshift cluster-bot etc...) there is no default value and Theia CPU limit won't be set. As a consequence Theia is able to consume all the CPU that's available on a cluster node. But in real world clusters (with limitranges and quotas for namespaces) the default can be a value that makes Theia bootstrapping flaky and 3x slower. That's been observed on devsandbox and che-dev where the default cpu limit is set to 0.4 per container.

On the other hand, for clusters where the cpu limit quota is low (under 2 cores), workspaces may fail to start because Kubernetes will refuse to schedule it. But on such clusters workspace would be flaky anyway so refusing to schedule it is probably the right choice.

That said I am +1 for this change if we verify that we can make sure that there are no regressions when starting our most beefy stack (java-mysql) on the following platforms:

1. kind
2. minikube
3. crc
4. openshift dev-cluster
5. RHPDS
6. che.osio

tsmaeder (changes requested):
This change breaks my self-hosting workspace on both crc and docker-desktop with the message:

```
Error: Failed to run the workspace: "Unrecoverable event occurred: 'FailedScheduling', '0/1 nodes are available: 1 Insufficient cpu.', 'workspacen9z2ri5pmfxmmja1.workspace-768874d7c5-bh6gq'"
```
So we do have a regression as stated by @l0rd Blocking the PR until we have resolved what the way forward is.

ibuziuk:
@tsmaeder could you clarify what is your crc setup in terms of RAM and CPU?

tsmaeder:
@ibuziuk I never change the defaults, so whatever that is.
For docker-desktop, I have 7 CPU's and 15 Gigs or RAM. 
In other news: a fresh java-maven workspace does not come up either on crc with the same error message (when using the che-theia from this PR).

ibuziuk:
yeah, but how do you start crc e.g. `crc start -m 24576 -c 6` ?

tsmaeder:
The VM settings say 4 CPU's and 16Gb of RAM.

tsmaeder:
> theia is not designed to take advantage of multiple cores and I believe the best practice will be to keep CPU at 1 or below

Actually, there are at least two processes running in the theia container: the theia back-end and the plugin host. Also, whatever processes the built-in plugins are starting. Also, there is the possibility to run plugins in the theia container if we don't specify a sidecar image, which extends the likelihood of more processes being run in the theia container.
So there _is_ a benefit to having > 1 CPU available.

ibuziuk:
> Actually, there are at least two processes running in the theia container: the theia back-end and the plugin host. Also, whatever processes the built-in plugins are starting. Also, there is the possibility to run plugins in the theia container if we don't specify a sidecar image, which extends the likelihood of more processes being run in the theia container.
So there is a benefit to having > 1 CPU available.

@ericwil @tsmaeder l In this case, we need the plugins team to identify the right CPU limit for che-theia if 1000m does not work in the general case for self-hosting, and at the same time there is a benifit to having multicores

tsmaeder:
Also, shouldn't we have separate CPU request and limit settings to properly address the problem? If we want to ensure a workspace has enough resources, we should set a request, not a limit, right?

ibuziuk:
That is a good point and I believe both need to be set for a proper configuration:

- request -> smth. that is guaranteed by k8s
- limit -> make sure that container never go above the value

What I see in the repo is that only `memoryLimit` are set for the plugins (memoryRequest, cpuRequest and cpuLimit are missing).

tsmaeder:
I'm not sure what fields the meta.yaml supports currently. The documentation in the readme on https://github.com/tsmaeder/che-plugin-registry/ seems to be woefully out of date: even "cpuLimit" is not mentioned. 🤷

vitaliy-guliy:
I cannot start a workspace on my local Che minikube deployment with this minimal devfile 

```
---
apiVersion: 1.0.0
metadata:
  generateName: test-cpulimit
projects:
  -
    name: nodejs-mongodb-sample
    source:
      type: git
      location: "https://github.com/che-samples/nodejs-mongodb-sample"
components:
  - type: cheEditor
    reference: >-
      https://raw.githubusercontent.com/eclipse/che-plugin-registry/ibuziuk-patch-1/v3/plugins/eclipse/che-theia/next/meta.yaml
    alias: che-theia
```

Screenshot from 2020-11-27 12-43-26

On Hosted Che it seems to be working.

l0rd:
I cannot start the workspace on openshift dev-cluster neither where worker nodes have 1500m available:

```
k get nodes -o json | jq '.items[] | select( .metadata.labels["node-role.kubernetes.io/worker"] != null ) | .status.allocatable.cpu'
"1500m"
"1500m"
"1500m"
```

ibuziuk:
> I'm not sure what fields the meta.yaml supports currently. The documentation in the readme on https://github.com/tsmaeder/che-plugin-registry/ seems to be woefully out of date: even "cpuLimit" is not mentioned.

@tsmaeder your fork is outdated ;-) see https://github.com/eclipse/che-plugin-registry#plugin-meta-yaml-structure

ibuziuk:
@l0rd openshift dev-cluster = cluster bot?

v3/plugins/eclipse/che-theia/next/meta.yaml
```diff
@@ -89,6 +89,8 @@ spec:
          - exposedPort: 13132
          - exposedPort: 13133
      memoryLimit: "512M"
+     cpuLimit: "1000m"
```
l0rd, line 92: @ibuziuk looks like the problem on my side was that "1000m" got translated in "1k". It works if I set "1" instead.

v3/plugins/eclipse/che-theia/next/meta.yaml
```diff
@@ -89,6 +89,8 @@ spec:
...
+     cpuRequest: "500m"
```
l0rd, line 93: This is probably too high. Normally theia container uses 30m. If we set 500 as cpu request we are going to prevent other processes to use the rest of the memory, even if Theia is inactive, because we have blocked 500m for theia bootstrap.

v3/plugins/eclipse/che-theia/next/meta.yaml
ibuziuk: sure I can change, but as I understand for you it worked according to the https://github.com/eclipse/che/issues/18472

```
apiVersion: 1.0.0
metadata:
  name: alotofcpu
components:
  - cpuLimit: 1500m
    id: eclipse/che-theia/next
    type: cheEditor
```

v3/plugins/eclipse/che-theia/next/meta.yaml
ibuziuk: well, we can stick to `0.03` and `1`, but my understanding that the request is smth. that is guaranteed by k8s and since currently che-theia IDE is the most important container in the workspace pod, we might like to allocate more to guarantee the
proper functionality and prevent potential throttling

v3/plugins/eclipse/che-theia/next/meta.yaml
ibuziuk: according to the docs `m` seems to be the only parameter that is supported. @ericwill any input on this? 

> Kubernetes/OpenShift-spec CPU request string (e.g. "125m"). Refer to https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-cpu for details

tsmaeder:
> 
> 
> I'm not sure what fields the meta.yaml supports currently. The documentation in the readme on https://github.com/tsmaeder/che-plugin-registry/ seems to be woefully out of date: even "cpuLimit" is not mentioned. 🤷

If I stopped looking at my way-out-of-date fork instead of https://github.com/eclipse/che-plugin-registry/, I might actually see the light 🤦

v3/plugins/eclipse/che-theia/next/meta.yaml
```diff
@@ -89,6 +89,8 @@ spec:
          - exposedPort: 13132
          - exposedPort: 13133
      memoryLimit: "512M"
+     cpuLimit: "1000m"
+     cpuRequest: "30m"
```
tsmaeder, line 93: Not sure we should run theia with 1/3rd of a single CPU.

v3/plugins/eclipse/che-theia/next/meta.yaml
tsmaeder, line 92: As commented elsewhere....we don't want to limit theia to 1 CPU either.

ibuziuk: are you absolutely sure that theia is designed to be multicore supported?

v3/plugins/eclipse/che-theia/next/meta.yaml
ibuziuk: hmm.. not sure I understand is not 1/3rd it is about requesting only 3% of the core with a 1 full core limit

v3/plugins/eclipse/che-theia/next/meta.yaml
tsmaeder: 3%, even worse....no useful work will happen with 3% of a cpu.

benoitf:
I think the values should be qualified (instead of doing like failed attempts) and probably it's not gonna work with absolute values for all use-cases so we might end with another mechanism

v3/plugins/eclipse/che-theia/next/meta.yaml
l0rd: It's even less than that. It's 0.03 of a CPU.

v3/plugins/eclipse/che-theia/next/meta.yaml
vitaliy-guliy: completely agree with @tsmaeder that 3% is too small

ibuziuk:
@benoitf I tried to qualified it and do want feedback from the plugins team in this regard:

- setting `1000m` limit since my understanding that theia is not designed to take advantage of multiple cores and I the best practice in general is to keep CPU at 1 or below
- settting `30m` for request based on @l0rd comment https://github.com/eclipse/che-plugin-registry/pull/716#discussion_r531605123

If this setting does not work for whatever reason, I would request the plugin team to provide a proper setup based on their expirience.

tsmaeder:
If we _need_ to specify a cpu limit, we should select a value that is big enough that we don't profit from making the value bigger. After all, the limit only is relevant if the container actually _uses_ the cpu, right? So if we specify cpuLimit=10, it doesn't really matter if theia only uses 1 cpu.
We also should put in a cpuRequest=x where x is a realistic value that will yield a halfway useful user experience. I'm thinking of something like 0.7.

v3/plugins/eclipse/che-theia/next/meta.yaml
ibuziuk: @tsmaeder @vitaliy-guliy do you folks understand the difference between a request & limit?

ericwill:
A few things:
* I don't think we should merge this today, there are too many variables and regressions possible. We need time to properly investigate what a reasonable limit would be, and its effects on all the different workflows.
* I'm thinking this fix does not belong in upstream Che...if the issue is for CRW then we should probably apply the limit only to the CRW plugin registry `meta.yaml`. This way we can keep the change surgical, as we have more control over the use cases in CRW than we do in upstream Che.

Bottom line: we need more time to investigate this properly.

v3/plugins/eclipse/che-theia/next/meta.yaml
tsmaeder: > Normally theia container uses 30m

Where do you get this number from? IMO, the theia back-end uses more than 3% of a single core as soon as you're doing something.

ibuziuk:
> Bottom line: we need more time to investigate this properly.

@ericwill ok, up to you folks - will let you do the calculation 

@tsmaeder I think you are probably missing what request & limits are expected to do.

v3/plugins/eclipse/che-theia/next/meta.yaml
tsmaeder: Yes, but if the scheduler sees a 3% request, it might schedule > 30 of those containers for a single cpu, I doubt that we want that.

l0rd:
> this fix does not belong in upstream Che

@ericwill why?

v3/plugins/eclipse/che-theia/next/meta.yaml
l0rd: @tsmaeder we are currently failing to schedule even one unique pod and you worry that we "might" schedule 30? Schedulers consider cpu limits as well and won't let you schedule pods if the max cpu limit is exceeded (as you have experienced on crc).

l0rd:
Anyway @ericwill I agree with you on the "not ready to merge" part. Validations have failed on most of the platforms.

ericwill:
> > this fix does not belong in upstream Che
> 
> @ericwill why?

It's not a concrete statement, it could still go either way (I'm mostly thinking out loud). My reasoning is:
* The use cases of upstream Che are so varied that it will be hard to pin down an exact limit that works. For example, in upstream we have double (if not more) the amount of plugins than in CRW, so there are many more different combinations of workflows that could be broken. It will be hard to test all of them.
* The nature of the workflows is different: upstream we are doing Che in Che development, among other things. I doubt many customers, if any, are doing this type of work. Their workflows, as it concerns theia's CPU usage, are going to be more routine.
* Devsandbox, AFAIUI, is based on CRW. If we can make the fix more surgical then we may as well, at least until some upstream user reports this use case.

v3/plugins/eclipse/che-theia/next/meta.yaml
tsmaeder: That may be, but it's has nothing to do with the question what the right cpuRequest value is.. I still don't want to expose our users to a theia that runs on 3% of a single core.

l0rd: It won't run on 3% of a single core. 3% will be reserved but the container will be able to use up to the CPU limit. What's the right number for you? Consider that currently the default CPU request for che-theia is 0%. We are exposing our users to a theia that reserves 0% of CPU.

l0rd:
> It's not a concrete statement, it could still go either way (I'm mostly thinking out loud). My reasoning is:
> 
> * The use cases of upstream Che are so varied that it will be hard to pin down an exact limit that works. For example, in upstream we have double (if not more) the amount of plugins than in CRW, so there are many more different combinations of workflows that could be broken. It will be hard to test all of them.

On most of the clusters out there the default CPU limit is not enough to run Theia. Specifying a cpu limit is a good practice that applies to both Che and CRW.

> * The nature of the workflows is different: upstream we are doing Che in Che development, among other things. I doubt many customers, if any, are doing this type of work. Their workflows, as it concerns theia's CPU usage, are going to be more routine.

CPU limits can be overridden by the values set on the devfile. That's how you would address that if you . And Che in Che development is something that we should be able to do on devsandbox as well. We should think about devsandbox as gitpod on OpenShift.

> * Devsandbox, AFAIUI, is based on CRW. If we can make the fix more surgical then we may as well, at least until some upstream user reports this use case.

I am +1 for surgical changes but those should be done upstream. We work upstream first. It's not effective to maintain 2 (3 with sandbox?) different code bases.

l0rd:
> I think the values should be qualified (instead of doing like failed attempts) and probably it's not gonna work with absolute values for all use-cases so we might end with another mechanism

@benoitf right that's why I would like to verify that it works with the defaults of the most used platforms. We already set memory requests/limits, we should be able to set cpu as well. 

There is something that we are missing because it should work on minikube/crc. And I would encourage @ibuziuk to investigate on why minikube/crc/cluster-bot are not working. Not setting the CPU request was the problem at the beginning because, if not set, the CPU request is set to match the CPU limit and that make the pod unschedulable.

Anyway on the mid-term we should do the following things to address this problem properly:

- Profile che-theia CPU/memory usage
- Profile extensions that are included in Che-Theia
- Profile plugins that are not included in Che-Theia but that has not a container specified in their meta.yaml
- Update the plugin broker to sum up all the CPU/memory limits/requests of all the plugins running in the same container

There was also some talking about making Theia cpu request/limits configurable as a new che-server property. I am not sure this is really necessary and I would prefer to not set have a property dedicated to a specific editor (che-theia). But that's something to think about.

v3/plugins/eclipse/che-theia/next/meta.yaml
l0rd: @ibuziuk yes specifying 1500m worked on the devfile but I was wondering if on the meta.yaml the value is parsed/converted in another part of the code. I will try to reproduce it on my local minikube.

l0rd:
> @l0rd openshift dev-cluster = cluster bot?

@ibuziuk it should be the same configuration yes. It's just that dev-cluster last 11 hours vs 2 hours of cluster-bot.

ibuziuk:
@l0rd I reproduced the issue against cluster bot (3 workers - 3500m each) `0/6 nodes are available: 6 Insufficient cpu.` 

The reason is that `30m` is translated to 30 cores, and this is definitely a bug on our end:


```
    Requests:
      cpu:     30

```

Another unrelated observation: jwt proxy is the only container that specifies CPU limit atm which is pretty high:

```
    Limits:
      cpu:     500m
      memory:  134217728
    Requests:
      cpu:     500m
      memory:  134217728
```

According to the 3.11 graphs request should be fairly low -  0.005:

image

ibuziuk:
> @ericwill in general we should have some really good reason to not work upstream first.

:+1: everything we do should land in the upstream first and each plugin should specify:

- memoryRequest
- memorylimit 
- cpuRequest 
- cpuLimit 

Opened https://github.com/eclipse/che/issues/18485

benoitf:
> The reason is that 30m is translated to 30 cores, and this is definitely a bug on our end:

@ibuziuk is there a bug for this ?

ibuziuk:
@ericwill if I set the values without `m` this results in the schema error - ```The command '/bin/sh -c ./check_metas_schema.sh v3' returned a non-zero code: 1
Error: Process completed with exit code 1.```

ericwill:
Superseded by #796

PR #757, opened 2020-12-10
add lombok extension

gattytto:
Closes: https://github.com/eclipse/che/issues/17483


che-theia-plugins.yaml
```diff
@@ -547,3 +547,15 @@ plugins:
       revision: v5.8.0
     extensions:
       - https://open-vsx.org/api/esbenp/prettier-vscode/5.8.0/file/esbenp.prettier-vscode-5.8.0.vsix
+  - repository:
+      url: https://github.com/gabrielbb/vscode-lombok
+      revision: v1.0.2
```
ericwill, line 552: I don't see this revision in the repo, maybe we should just use master?

PR #761, opened 2020-12-12
add lombok extension

gattytto:
@ericwill maybe better have che-theia-plugins.d/extension-name.yaml instead? 
Closes: https://github.com/eclipse/che/issues/17483

### DIFF
--- a/che-theia-plugins.yaml
+++ b/che-theia-plugins.yaml
@@ -562,3 +562,15 @@ plugins:
       revision: 1e15d3495da89072d48cf583d48d92829f0c4b82
     extensions:
       - https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-eslint/vscode-eslint-2.1.1-1e15d3.vsix
+  - repository:
+      url: https://github.com/gabrielbb/vscode-lombok
+      revision: master
+    sidecar:
+      directory: java8
+      name: vscode-lombok
+      memoryLimit: "1500Mi"
+      volumeMounts:
+      - path: "/home/theia/.m2"
+        name: m2
+    extensions:
+      - https://open-vsx.org/api/cool2k/vscode-lombok/1.0.2/file/cool2k.vscode-lombok-1.0.2.vsix
--- a/che-theia-plugins.yaml
+++ b/che-theia-plugins.yaml
@@ -566,7 +566,7 @@ plugins:
       url: https://github.com/gabrielbb/vscode-lombok
       revision: master
     sidecar:
-      directory: java8
+      directory: java
       name: vscode-lombok
       memoryLimit: "1500Mi"
       volumeMounts:


che-theia-plugins.yaml
```diff
@@ -562,3 +562,15 @@ plugins:
       revision: 1e15d3495da89072d48cf583d48d92829f0c4b82
     extensions:
       - https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-eslint/vscode-eslint-2.1.1-1e15d3.vsix
+  - repository:
+      url: https://github.com/gabrielbb/vscode-lombok
+      revision: master
+    sidecar:
+      directory: java8
+      name: vscode-lombok
```
benoitf, line 570: does it works with java (java11) ? As it's the latest lts

gattytto: yes and I made the corresponding changes to demo, devfile and extension config

gattytto:
image

sunix (changes requested):
Hello,
As mentionned in https://github.com/eclipse/che-devfile-registry/pull/316/files#r544386479, the plugin won't work out of the box if the lombok.jar is not copied to `/tmp/vscode-unpacked`. We cannot ask the user who would compose its devfile to magically know that something like that has to be copied from the devfile.
So first, does it means that we have something wrong in Che-theia ? why lombok.jar can't be found in the original location.
Then is the workaround to do the copy from the plugin sidecar by overriding the entrypoint ?

Last, I don't think it is a good idea to keep Java and Lombok into 2 seperate plugins: we would need to use mergePlugins (that was not working for me BTW) or share additional volumes like you did in your devfile. Again it is not straight forward for the user.

sunix:
Tested with 
```yaml
metadata:
  generateName: java-maven-lombok-
attributes:
  mergePlugins: 'true'
projects:
  - name: lombok-project-example
    source:
      location: 'https://github.com/chuucks/LOMBOK-PROJECT-EXAMPLE'
      type: git
      branch: master
components:
  - id: cool2k/vscode-lombok/latest
    preferences:
      java.server.launchMode: Standard
    type: chePlugin
    registryUrl: 'https://sutan-che-plugin-registry-dev-2vqkx.surge.sh/v3/'
  - mountSources: true
    endpoints:
      - name: debug
        port: 5005
        attributes:
          public: 'false'
    memoryLimit: 512Mi
    type: dockerimage
    volumes:
      - name: m2
        containerPath: /home/user/.m2
    image: 'quay.io/eclipse/che-java11-maven:7.22.1'
    alias: maven
    env:
      - value: ''
        name: MAVEN_CONFIG
      - value: >-
          -XX:MaxRAMPercentage=50 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
          -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4
          -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true
          -Xms20m -Djava.security.egd=file:/dev/./urandom -Duser.home=/home/user
        name: MAVEN_OPTS
      - value: >-
          -XX:MaxRAMPercentage=50 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
          -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4
          -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true
          -Xms20m -Djava.security.egd=file:/dev/./urandom
        name: JAVA_OPTS
      - value: >-
          -XX:MaxRAMPercentage=50 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
          -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4
          -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true
          -Xms20m -Djava.security.egd=file:/dev/./urandom
        name: JAVA_TOOL_OPTIONS
apiVersion: 1.0.0
commands:
  - name: maven build
    actions:
      - workdir: '${CHE_PROJECTS_ROOT}/lombok-project-example'
        type: exec
        command: mvn clean install
        component: maven

```

sunix:
I created that issue https://github.com/eclipse/che/issues/18632 to see if this is a bug or not.

sunix:
The lombok extension is
- providing `lombok.jar`
- updating the user settings `java.jdt.ls.vmargs` with the `lombok.jar` location

To me it is not an issue to add a volume `/tmp/vscode-unpacked` to all the java plugins containing the vscode-java extension (quarkus plugin, java plugin, java8 plugin, etc ..) + the lombok plugin.
```
      volumes:
        - name: java-vscode-unpacked
          mountPath: /tmp/vscode-unpacked
```
That way, the vscode-java extension would have access to the `lombok.jar` provided by the vscode-lombok extension.

Having said that, we have a blocker issue:
> The lombok extension is
> - providing `lombok.jar`
> - updating the user settings `java.jdt.ls.vmargs` with the `lombok.jar` location

The problem is that the user settings are reused by any workspaces started by the same user. So if the user starts a new workspace with Java plugin without the lombok extension ... it would crash not finding the lombok.jar.
Looks like this is related to https://github.com/GabrielBB/vscode-lombok/issues/49

Several proposals we could choose to solve that
- Propose a patch to the current lombok extension adding a preference option to use workspace settings instead of user settings.
- Create our own extension to be included in our all java plugins. This would have a preference option to activate lombok: in the workspace settings updating `java.jdt.ls.vmargs` with the `lombok.jar` location that would also be provided by the extension.
- Propose a patch to the current lombok extension adding a preference option to choose the location of jar. The extension should copy the jar in that location and update the user settings. But that would requires that we have a volume available in all the user workspaces. the same `home` folder available to the any workspaces of the same user. Could be advanced implementation of https://github.com/eclipse/che/issues/15421

sunix:
@gattytto FYI, I am closing this PR in favor of https://github.com/eclipse/che-plugin-registry/pull/860

PR #783, opened 2020-12-23
Plantuml plugin

gattytto:
Closes: https://github.com/eclipse/che/issues/17916


gattytto:
tested OK

sidecars/plantuml/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+FROM plantuml/plantuml-server
+
+ENV TMPDIR=/tmp
+ENV HOME=/home/theia
+
+ADD etc/storage.conf $HOME/.config/containers/storage.conf
```
sunix, line 15: what is that storage.conf for ?

che-theia-plugins.yaml
```diff
@@ -1,5 +1,16 @@
 version: 1.0.0
 plugins:
+  - repository:
```
sunix, line 3: The plugins part needs to be in a separate PR (the sidecar that would be generated not pushed to quay.io yet) see https://github.com/eclipse/che-plugin-registry/blob/master/CONTRIBUTE.md#contribution-flow

sidecars/plantuml/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+FROM plantuml/plantuml-server
+
+ENV TMPDIR=/tmp
```
sunix, line 12: Is this mandatory?

sunix (changes requested):

che-theia-plugins.yaml
ericwill: You can actually merge them together if you really want to, but testing this PR will be difficult because the generated surge link won't contain a valid sidecar reference. :man_shrugging:

gattytto:
closing to refactor with new plungin-registry scheme

PR #797, opened 2021-01-14
Add support for building with buildah

ericwill:

### What does this PR do?
Not all users have access to docker, so provide the --buildah option to build the registry with buildah.

Part of eclipse/che#18798

### How to test this PR?
Run `build.sh --buildah` and verify that the registry was built successfully.

### PR Checklist

- [x] [Tests are covering the bugfix](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#tests-are-covering-the-bugfix)
- [x] [Relevant user documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [Relevant contributing documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [CI/CD changes implemented, documented and communicated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#cicd-changes-implemented-documented-and-communicated)

### Reviewers


build.sh
```diff
@@ -90,21 +97,42 @@ VERSION=$(head -n 1 VERSION)
 case $VERSION in
   *SNAPSHOT)
     echo "Snapshot version (${VERSION}) specified in $(find . -name VERSION): building nightly plugin registry."
-    docker build \
-        -t "${IMAGE}" \
-        -f ${DOCKERFILE} \
-        --build-arg LATEST_ONLY="${LATEST_ONLY}" \
-        --build-arg "USE_DIGESTS=${USE_DIGESTS}" \
-        --target "${TARGET}" .
+    if [ $USE_BUILDAH = true ]
```
benoitf, line 100: In some other repo we try to detect with `which` command if docker or podman commands are available so no need to use switch

ericwill: what if both are installed, but the user wants to use a specific one?

benoitf: if it's building I would say users may not care but you could keep flag if you think it's important

I think that it's nice if it's building trying available tools without knowing special flags

ericwill: I am thinking more of the case where someone's machine has docker pre-installed but without root access, in which case they'd want to force the usage of buildah

nickboldt: if there's a reason to explicitly ask for podman/buildah when docker is also present, then this approach makes sense.

In other places, we have stuff like:

```
PODMAN=$(command -v podman || true)
if [[ ! -x $PODMAN ]]; then
  echo "[WARNING] podman is not installed."
 PODMAN=$(command -v docker || true)
  if [[ ! -x $PODMAN ]]; then
    echo "[ERROR] docker is not installed. Aborting."; exit 1
  fi
fi
```

https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/crw-2.6-rhel-8/build.sh#L46-L53

ericwill: I forgot about `podman`, I'll adjust the script

ericwill:
> looks odd to have PODMAN variable set to docker ?

Any further comments @benoitf ?

benoitf:
@ericwill are we doing if else for buildah just to replace 'build' by 'bud' ? If so can't we just replace it with another variable ?

ericwill:
Seems that `docker` is a hardcoded step in the build PR check. @benoitf maybe we should use docker by default, to keep the CI happy?

nickboldt:
unless you need `docker buildx` I'd say use podman in favour of docker. But that might just be personal preference :)

ericwill:
> unless you need `docker buildx` I'd say use podman in favour of docker. But that might just be personal preference :)

`podman` has no multi-arch at all, so using podman in the CI would effectively close the door on such work :frowning_face:

benoitf:
we're moving to buildx but what you can do is that if builder is docker and buildx is enablee we enable buildx parameter else we use podman/buildah

But yeah PR check need to validate all archs

ericwill:
> we're moving to buildx but what you can do is that if builder is docker and buildx is enablee we enable buildx parameter else we use podman/buildah
> 
> But yeah PR check need to validate all archs

Another option is to add `BUILDER=docker` in all invocations of `buildh.sh` in the CI

benoitf:
yes fine as well

.github/workflows/image-build-pr-check.yml
```diff
@@ -56,10 +56,10 @@ jobs:
         export SKIP_TEST=true
         if [[ "${{matrix.dist}}" == "alpine" ]]; then
           docker pull quay.io/eclipse/che-plugin-registry:nightly
-          ./build.sh --tag pr-check
+          BUILDER=docker ./build.sh --tag pr-check
```
benoitf, line 59: Might be interesting to have in Matrix job a build with podman and buildah as well for PR checks

ericwill: `docker` is used to run the containers, so we'd have to run them with `podman`, which seems a bit complicated for a PR check. It's unlikely that something will build on `docker` but fail with `podman`.

ericwill: Let's explore it another time. Okay to merge this PR?

benoitf: sure 
Fine to explore it another time

PR #861, opened 2021-02-25
add UML extension

gattytto:
up for review with new schema.
Closes: https://github.com/eclipse/che/issues/17916


che-plugins.yaml
```diff
@@ -172,3 +172,14 @@ plugins:
...
             path: /var/lib/buildkit
+  - repository:
+      url: https://github.com/qjebbs/vscode-plantuml
+      revision: v2.14.0
+    sidecar:
+      directory: plantuml
+      name: vscode-plantuml
+      memoryLimit: '512Mi'
+    aliases:
+      - qjebbs/vscode-plantuml
```
ericwill, line 183: We don't need an alias unless there are multiple versions of the same plugin, from the same repository. So this line is not needed.

sidecars/plantuml/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2019 Red Hat, Inc.
```
ericwill, line 1: This needs to be in a separate PR first

PR #926, opened 2021-04-14
The "che-plugin-registry" openshift PR check job fails because unable to login

Ohrimenko1988:
Signed-off-by: Ihor Okhrimenko <iokhrime@redhat.com>

### What does this PR do?
Add user provisioning and pass necessary env variables to the test container

Issue: https://github.com/eclipse/che/issues/19588
### PR Checklist

- [ ] [The Eclipse Contributor Agreement is valid](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-eclipse-contributor-agreement-is-valid)
- [ ] [Code produced is complete](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-produced-is-complete)
- [ ] [Code builds without errors](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-builds-without-errors)
- [ ] [Sections `What issues does this PR fix or reference` and `How to test this PR` completed](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#sections-what-issues-does-this-pr-fix-or-reference-and-how-to-test-this-pr-completed)

### Reviewers


.ci/openshift-ci/users.htpasswd
```diff
@@ -0,0 +1 @@
+user:{SHA}Et6pb+wgWTVmq3VpLJlJWWgzrck=
```
benoitf, line 1: if it's a dummy credential like 'user' it should be generated by the script and not stored in the repository
if it's a real credential, it needs to be stored with github secrets.

.ci/openshift-ci/test.sh
```diff
@@ -48,7 +65,7 @@ END
 }
 
 deployChe() {
-  chectl server:deploy  --che-operator-cr-patch-yaml=custom-resources.yaml --telemetry=off --workspace-engine=dev-workspace --platform=openshift --installer=operator --batch
+  chectl server:deploy  --che-operator-cr-patch-yaml=custom-resources.yaml --telemetry=off --platform=openshift --installer=operator --batch
```
benoitf, line 68: telemetry should be turned off by config like

```bash
  # turn off telemetry
  mkdir -p ${HOME}/.config/chectl
  echo "{\"segment.telemetry\":\"off\"}" > ${HOME}/.config/chectl/config.json
```

so we don't need to append --telemetry=off on every command

benoitf, line 68: double space between deploy and --che

.ci/openshift-ci/happy-path-pod.yaml
```diff
@@ -12,6 +12,18 @@ spec:
       image: quay.io/eclipse/che-e2e:nightly
       imagePullPolicy: Always
       env:
+        - name: TS_SELENIUM_OCP_USERNAME
```
benoitf, line 15: this file shouldn't be committed at all in che-plugin-registry repository

benoitf:
AFAIK these scripts are highly tied to happy path and not to plug-in registry, then most of them should be part of che/tests/e2e and just launched here from this repository

.ci/openshift-ci/test.sh
Ohrimenko1988: The "chectl" used only in one place so in my eyes using additional "--telemetry=off" flag looks more clear than
```
  # turn off telemetry
  mkdir -p ${HOME}/.config/chectl
  echo "{\"segment.telemetry\":\"off\"}" > ${HOME}/.config/chectl/config.json
  ```

File: .ci/openshift-ci/happy-path-pod.yaml (hunk 3 above)
Comment (Ohrimenko1988) [reply .ci/openshift-ci/happy-path-pod.yaml:15]: This file uses for setup plugins test.
For running tests on the openshift CI we need to configure a pod.
You can see that we get the "che-e2e" nightly test image
and define a list of specific variables necessary for
running exactly plugins test.

**Comment by Ohrimenko1988** [comment]
> AFAIK these scripts are highly tied to happy path and not to plug-in registry, then most of them should be part of che/tests/e2e and just launched here from this repository

@rhopp Already works on the creation of the common workflow for openshift CI but even in this case, it gives a chance to reduce a script (remove che installation and etc) but not get rid of them at all.

File: .ci/openshift-ci/happy-path-pod.yaml (hunk 3 above)
Comment (benoitf) [reply .ci/openshift-ci/happy-path-pod.yaml:15]: but we can't get it from external location ?
because there is no link to che-plugin-registry in that file, it's a pure che e2e test config file

File: .ci/openshift-ci/test.sh [hunk 4]
```diff
@@ -65,44 +66,44 @@ END
...
 deployChe() {
-  chectl server:deploy  --che-operator-cr-patch-yaml=custom-resources.yaml --telemetry=off --platform=openshift --installer=operator --batch
+  chectl server:deploy --che-operator-cr-patch-yaml=custom-resources.yaml --telemetry=off --platform=openshift --installer=operator --batch
 }
 
 patchTestPodConfig(){
-  # obtain the basic "happy-path" pod config
-  cat .ci/openshift-ci/happy-path-pod.yaml > happy-path-pod.yaml
+  # obtain the basic test pod config
+  cat .ci/openshift-ci/plugins-test-pod.yaml > plugins-test-pod.yaml
```
benoitf, line 74: could we just grab the pod definition from che e2e repository and just update config map with the tests to launch ?

.ci/openshift-ci/test.sh
```diff
@@ -65,44 +66,44 @@ END
...
 
   # just to sleep
   sleep 3
 
   # download the test results
   mkdir -p /tmp/e2e
-  oc rsync -n ${TEST_POD_NAMESPACE} ${HAPPY_PATH_POD_NAME}:/tmp/e2e/report/ /tmp/e2e -c download-reports
-  oc exec -n ${TEST_POD_NAMESPACE} ${HAPPY_PATH_POD_NAME} -c download-reports -- touch /tmp/done
+  oc rsync -n ${TEST_POD_NAMESPACE} ${TEST_PATH_POD_NAME}:/tmp/e2e/report/ /tmp/e2e -c download-reports
+  oc exec -n ${TEST_POD_NAMESPACE} ${TEST_PATH_POD_NAME} -c download-reports -- touch /tmp/done
```
benoitf, line 106: it looks like all this script could be in che e2e tests as the only 'plugin specific stuff' seems to be where we say: use 'test-all-plugins'

.ci/openshift-ci/test.sh
Ohrimenko1988: The "plugin-test-pod.yaml" has defined exactly for che-plugin-registry set of parameters.
As you can see we patch only product url which defined in runtime.

benoitf: but we don't need to have 64 lines of yaml just to change product URL
all the settings are default happy path configuration that are tied to che/e2e, not this repository

benoitf: it looks like script should be in https://github.com/eclipse/che/tree/master/tests/.infra or https://github.com/eclipse/che/tree/master/tests/e2e/files/happy-path

.ci/openshift-ci/test.sh
Ohrimenko1988: I am not sure that this discussion is a topic of this issue.
If your point to remove this script from the "che-plugin-registry" repository,
please create specific issue for that and we can continue the discussion.

.ci/openshift-ci/test.sh
Ohrimenko1988: We need it for defining all these parameters.
Besides, please pat attention that these tests have no relations to the "Happy-Path"
```
      env:
        - name: TS_SELENIUM_OCP_USERNAME
          value: 'user'
        - name: TS_SELENIUM_OCP_PASSWORD
          value: 'user'
        - name: TS_SELENIUM_VALUE_OPENSHIFT_OAUTH
          value: 'true'
        - name: TS_SELENIUM_USERNAME
          value: 'admin'
        - name: TS_SELENIUM_PASSWORD
          value: 'admin'
        - name: TS_OCP_LOGIN_PAGE_PROVIDER_TITLE
          value: 'htpasswd'
        - name: TEST_SUITE
          value: test-all-plugins
        - name: NODE_TLS_REJECT_UNAUTHORIZED
          value: '0'
        - name: TS_SELENIUM_BASE_URL
          value: CHE_URL
        - name: TS_SELENIUM_LOG_LEVEL
          value: DEBUG
        - name: TS_SELENIUM_MULTIUSER
          value: 'true'
        - name: TS_SELENIUM_VALUE_TLS_SUPPORT
          value: 'true'
        - name: TS_SELENIUM_W3C_CHROME_OPTION
          value: 'true'          
      volumeMounts:
        - name: test-run-results
          mountPath: /tmp/e2e/report/
      resources:
        requests:
          memory: "8Gi"
          cpu: "2000m"
        limits:
          memory: "8Gi"
          cpu: "4000m" 
```

.ci/openshift-ci/test.sh
benoitf: well it should not have been merged like that, we're creating technical debt

.ci/openshift-ci/test.sh
benoitf: but it looks like this one: https://github.com/eclipse/che/blob/f8a7aaef6b7db0b59bef382f5eefaac204236ca0/tests/performance/pod-oauth.yaml

nothing it tied to plugin registry in the values you reference

.ci/openshift-ci/test.sh
ericwill: We can always abstract them out at a later time, right now we're trying to get the tests to run properly first

.ci/openshift-ci/test.sh
Ohrimenko1988: The example you provided has a different set of parameters. Besides this file in the Che repo because the "load-tests" rely on che. Let me explain my opinion. I am trying to explain that file we discussing for is specific for che-plugin-registry , so I don't want to keep something specific for one repo in the common "e2e" repository. I'll prefer to store specific files in the specific repo.

benoitf: all the env value except `test-all-plugins` look like default properties when you want to test Eclipse Che.
so all these defaults should be in e2e/tests repository

it's like
```
volumeMounts:
        - name: test-run-results
          mountPath: /tmp/e2e/report/
```

it's not related to plug-in registry as well

.ci/openshift-ci/build/Dockerfile
```diff
@@ -32,9 +32,10 @@ RUN yum install nodejs -y
 RUN npm install --global yarn
 
 # Build repo
-RUN yarn && \
-    yarn --cwd tools/build build && \
-    eval node tools/build/lib/entrypoint.js --output-folder:"output"
+RUN yarn
+RUN yarn install
+RUN cd tools/build && yarn build
+RUN pwd && eval yarn node tools/build/lib/entrypoint.js --output-folder:"output"
```
vitaliy-guliy, line 38: Why not to leave it as one instruction?

Ohrimenko1988: It was split for separated rows because I faced a problem and it was necessary to perform a debug.
It will be cleaned up before merging.
Thank you for reminding me.

.ci/openshift-ci/build/Dockerfile
```diff
@@ -24,17 +24,17 @@ ADD . /che-plugin-registry
...
+    yum install openssh -y &&\
+    curl -sL https://rpm.nodesource.com/setup_14.x | bash - &&\
+    yum install nodejs -y &&\
+    npm install --global yarn
 
 # Build repo
-RUN yarn && \
-    yarn --cwd tools/build build && \
-    eval node tools/build/lib/entrypoint.js --output-folder:"output"
+RUN yarn && yarn install
```
ericwill, line 35: Wouldn't it be easier to run the `build.sh` script instead of the commands directly? You can run it so that it doesn't produce a build, i.e. `build.sh --skip-oci-image`

Ohrimenko1988: Good catch, thank you. Done.

.ci/openshift-ci/build/Dockerfile
```diff
@@ -24,17 +24,16 @@ ADD . /che-plugin-registry
...
+    curl -sL https://rpm.nodesource.com/setup_14.x | bash - &&\
+    yum install nodejs -y &&\
+    npm install --global yarn
 
 # Build repo
-RUN yarn && \
-    yarn --cwd tools/build build && \
-    eval node tools/build/lib/entrypoint.js --output-folder:"output"
+
+RUN ./build.sh --skip-oci-image
```
ericwill, line 36: Sorry I forgot to mention the other options which may be useful: `SKIP_TEST=true SKIP_FORMAT=true SKIP_LINT=true ./build.sh --skip-oci-image --skip-digest-generation`

.ci/openshift-ci/common-functions.sh
```diff
@@ -0,0 +1,213 @@
+#!/bin/bash
+
+#
+# Copyright (c) 2012-2021 Red Hat, Inc.
```
ericwill, line 4: Not sure if we need the 2012 in here.

build.sh
```diff
@@ -88,7 +88,7 @@ function parse_arguments() {
 parse_arguments "$@"
 
 echo "Update yarn dependencies..."
-yarn
+yarn
```
ericwill, line 91: Looks like a trailing space was introduced

ericwill (approved):
Approving for merging after minor fixes

PR #937, opened 2021-04-27
fix(entrypoint): Fix image regexp when using CHE_SIDECAR_CONTAINERS_REGISTRY_URL

sunix:
### What does this PR do?
1. Fix problems with image regexp:
   - single quotes not supported
   - multiline not supported
   - multiple occurence
2. Adding unit tests script

N/A

https://github.com/eclipse/che/issues/19405 and https://github.com/eclipse/che/issues/19492

### How to test this PR?
1. Build the registry from this branch:
   ```
   ./build.sh
   ```
3. run the built container and check that  meta.yaml generated are OK, for instance `/var/www/html/v3/plugins/ms-python/python/latest/meta.yaml`
   ```
   podman run -t quay.io/eclipse/che-plugin-registry:nightly cat /var/www/html/v3/plugins/ms-python/python/latest/meta.yaml
   ```
   should contain these lines
    ```yaml
      - image: 'quay.io/eclipse/che-plugin-sidecar@sha256:64311210e710ca1614839f897285fc1d39b2ca17215d56c165d091aafd19a14b'
   ```
4. check that meta.yaml works well with the env variable `CHE_SIDECAR_CONTAINERS_REGISTRY_URL`
   ```
   podman run -e CHE_SIDECAR_CONTAINERS_REGISTRY_URL=https://fakeregistry.io:5000 -t quay.io/eclipse/che-plugin-registry:nightly cat /var/www/html/v3/plugins/ms-python/python/latest/meta.yaml
   ```
   should contain these lines
    ```yaml
    - image: 'https://fakeregistry.io:5000/eclipse/che-plugin-sidecar@sha256:64311210e710ca1614839f897285fc1d39b2ca17215d56c165d091aafd19a14b'
   ```

### PR Checklist

- [x] [Tests are covering the bugfix](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#tests-are-covering-the-bugfix)

### Reviewers

### DIFF
--- a/build/dockerfiles/entrypoint.sh
+++ b/build/dockerfiles/entrypoint.sh
@@ -37,31 +37,6 @@ METAS_DIR="${METAS_DIR:-${DEFAULT_METAS_DIR}}"
 #   \7 - Optional quotation following image reference
 IMAGE_REGEX="([[:space:]>-]*[\r]?[[:space:]]*[\"']?)([._:a-zA-Z0-9-]*)/([._a-zA-Z0-9-]*)/([._a-zA-Z0-9-]*)(@sha256)?:([._a-zA-Z0-9-]*)([\"']?)"
 
-function update_container_image_references() {
-
-    # We can't use the `-d` option for readarray because
-    # registry.centos.org/centos/httpd-24-centos7 ships with Bash 4.2
-    # The below command will fail if any path contains whitespace
-    readarray -t metas < <(find "${METAS_DIR}" -name 'meta.yaml')
-    for meta in "${metas[@]}"; do
-    echo "Checking meta $meta"
-    # Need to update each field separately in case they are not defined.
-    # Defaults don't work because registry and tags may be different.
-    if [ -n "$REGISTRY" ]; then
-        echo "    Updating image registry to $REGISTRY"
-        < "$meta" tr '\n' '\r' | sed -E "s|image:$IMAGE_REGEX|image:\1${REGISTRY}/\3/\4\5:\6\7|" |  tr '\r' '\n' > "$meta.tmp" && cat "$meta.tmp" > "$meta" && rm "$meta.tmp"
-    fi
-    if [ -n "$ORGANIZATION" ]; then
-        echo "    Updating image organization to $ORGANIZATION"
-        < "$meta" tr '\n' '\r' | sed -E "s|image:$IMAGE_REGEX|image:\1\2/${ORGANIZATION}/\4\5:\6\7|" |  tr '\r' '\n' > "$meta.tmp" && cat "$meta.tmp" > "$meta" && rm "$meta.tmp"
-    fi
-    if [ -n "$TAG" ]; then
-        echo "    Updating image tag to $TAG"
-        < "$meta" tr '\n' '\r' | sed -E "s|image:$IMAGE_REGEX|image:\1\2/\3/\4:${TAG}\7|" |  tr '\r' '\n' > "$meta.tmp" && cat "$meta.tmp" > "$meta" && rm "$meta.tmp"
-    fi
-    done
-
-}
 
 function run_main() {
     # Extract and use env variables with image digest information.
@@ -133,6 +108,32 @@ function run_main() {
     update_container_image_references
 
     exec "${@}"
+
+}
+
+function update_container_image_references() {
+
+    # We can't use the `-d` option for readarray because
+    # registry.centos.org/centos/httpd-24-centos7 ships with Bash 4.2
+    # The below command will fail if any path contains whitespace
+    readarray -t metas < <(find "${METAS_DIR}" -name 'meta.yaml')
+    for meta in "${metas[@]}"; do
+    echo "Checking meta $meta"
+    # Need to update each field separately in case they are not defined.
+    # Defaults don't work because registry and tags may be different.
+    if [ -n "$REGISTRY" ]; then
+        echo "    Updating image registry to $REGISTRY"
+        < "$meta" tr '\n' '\r' | sed -E "s|image:$IMAGE_REGEX|image:\1${REGISTRY}/\3/\4\5:\6\7|" |  tr '\r' '\n' > "$meta.tmp" && cat "$meta.tmp" > "$meta" && rm "$meta.tmp"
+    fi
+    if [ -n "$ORGANIZATION" ]; then
+        echo "    Updating image organization to $ORGANIZATION"
+        < "$meta" tr '\n' '\r' | sed -E "s|image:$IMAGE_REGEX|image:\1\2/${ORGANIZATION}/\4\5:\6\7|" |  tr '\r' '\n' > "$meta.tmp" && cat "$meta.tmp" > "$meta" && rm "$meta.tmp"
+    fi
+    if [ -n "$TAG" ]; then
+        echo "    Updating image tag to $TAG"
+        < "$meta" tr '\n' '\r' | sed -E "s|image:$IMAGE_REGEX|image:\1\2/\3/\4:${TAG}\7|" |  tr '\r' '\n' > "$meta.tmp" && cat "$meta.tmp" > "$meta" && rm "$meta.tmp"
+    fi
+    done
 
 }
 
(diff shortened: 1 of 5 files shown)


sunix:
still need to automate test execution from gh-action

build/dockerfiles/entrypoint.sh
```diff
@@ -35,94 +35,110 @@ METAS_DIR="${METAS_DIR:-${DEFAULT_METAS_DIR}}"
...
+    # registry.centos.org/centos/httpd-24-centos7 ships with Bash 4.2
+    # The below command will fail if any path contains whitespace
+    readarray -t metas < <(find "${METAS_DIR}" -name 'meta.yaml')
+    for meta in "${metas[@]}"; do
+    echo "Checking meta $meta"
+    # Need to update each field separately in case they are not defined.
+    # Defaults don't work because registry and tags may be different.
+    if [ -n "$REGISTRY" ]; then
+        echo "    Updating image registry to $REGISTRY"
+        < "$meta" tr '\n' '\r' | sed -E "s|image:$IMAGE_REGEX|image:\1${REGISTRY}/\3/\4\5:\6\7|" |  tr '\r' '\n' > "$meta.tmp" && cat "$meta.tmp" > "$meta" && rm "$meta.tmp"
```
cccs-eric, line 126: This `sed` command will only replace the first occurrence.  Some YAML files contain more than one image.  For example, take a look at `/var/www/html/v3/plugins/eclipse/che-theia/latest/meta.yaml` once you are in the container (and do `export CHE_SIDECAR_CONTAINERS_REGISTRY_URL=myregistry.com` and execute `/usr/bin/entrypoint.sh`:
```yaml
  containers:
    - image: 'myregistry.com/eclipse/che-theia@sha256:69b7d27a9e9a4b46c2734d995456385bb0d7ab1022638d95ddaa5a5919ef43c1'
      env:
        - name: THEIA_PLUGINS
          value: 'local-dir:///plugins'
        - name: HOSTED_PLUGIN_HOSTNAME
          value: 0.0.0.0
        - name: HOSTED_PLUGIN_PORT
          value: '3130'
        - name: THEIA_HOST
          value: 127.0.0.1
      mountSources: true
      memoryLimit: 512M
      volumes:
        - name: plugins
          mountPath: /plugins
        - name: theia-local
          mountPath: /home/theia/.theia
      name: theia-ide
      ports:
        - exposedPort: 3100
        - exposedPort: 3130
        - exposedPort: 13131
        - exposedPort: 13132
        - exposedPort: 13133
    - image: 'quay.io/eclipse/che-machine-exec@sha256:98fdc3f341ed683dc0f07176729c887f2b965bade9c27d16dc0e05f9034e624c'
      command:
        - /go/bin/che-machine-exec
        - '--url'
        - '127.0.0.1:3333'
```
Only the first reference is being replaced.  You could fix this by adding a global switch to your `sed` command:
```suggestion
        < "$meta" tr '\n' '\r' | sed -E "s|image:$IMAGE_REGEX|image:\1${REGISTRY}/\3/\4\5:\6\7|g" |  tr '\r' '\n' > "$meta.tmp" && cat "$meta.tmp" > "$meta" && rm "$meta.tmp"
```

I welcome the addition of test cases since those things should be caught at build time.  But I'm questioning the patch about temporarily replacing newlines (\n) by carriage returns (\r).  To me we are applying a band-aid solution to the problem instead of fixing the cause as I explained [here](https://github.com/eclipse/che/issues/19405#issuecomment-823373738).  I understand that some fields could be long (like `description`) and could be better written as multi-line, but are those YAML really meant to be read by humans?  Changing the YAML generation would avoid having to do tricky translations like we are doing now.  The fact that an `image: ` tag gets written as a multi-line field doesn't make sense to me.

This is my opinion and I'm totally fine if you disagree. 😃 No hard feelings!

benoitf:
I would either use `yq` or drop multiline on generation side rather than using complex sed expressions that may still fail.
We can easily remove multi-line as everything is generated so I would choose this option

About quotes we can't remove them as they're required due do the `:` character

build/dockerfiles/entrypoint.sh
sunix: > I would either use `yq` or drop multiline on generation side rather than using complex sed expressions that may still fail.
> We can easily remove multi-line as everything is generated so I would choose this option
> 
> About quotes we can't remove them as they're required due do the `:` character

@benoitf I would use `yq` but it should be part of a separate improvement issue. this PR is a fix on something that already exist. Would creating a new issue to use yq and prioritize it work for you?

sunix:
> I would either use `yq` or drop multiline on generation side rather than using complex sed expressions that may still fail.
> We can easily remove multi-line as everything is generated so I would choose this option
> 
> About quotes we can't remove them as they're required due do the `:` character

@benoitf I would use `yq` but it should be part of a separate improvement issue. this PR is a fix on something that already exist. Would creating a new issue to use yq and prioritize it work for you?

build/dockerfiles/entrypoint.sh
sunix: Thanks @cccs-eric,
I will add a test case for multiple occurrences.
About generating things differently, it would be less robust and may break at any time if we modify the generator. I would rather use yq but would like to have this part of a separate issue as it is a bit more complex (one more dependency and make sure we don't have regressions)

benoitf:
@sunix: I would prefer to just enforce that yaml are not multiline instead of trying to solve parsing with sed in the script
then probably we

ericwill:
> @sunix: I would prefer to just enforce that yaml are not multiline instead of trying to solve parsing with sed in the runtime script

+1 from me. I'm also in favour of using `yq` now, even if it takes a bit more time. Using `sed` for yaml is quite painful.

sunix:
> > @sunix: I would prefer to just enforce that yaml are not multiline instead of trying to solve parsing with sed in the runtime script
> 
> +1 from me. I'm also in favour of using `yq` now, even if it takes a bit more time. Using `sed` for yaml is quite painful.

@ericwill `yq` makes sense for the multi line and quotes problems, but at the end we would still need a regexp to extract/replace elements of the `image field`. not sure `yq` is good enough.

build/dockerfiles/test_entrypoint.sh
```diff
@@ -0,0 +1,350 @@
...
+}
+trap cleanup EXIT
+
+RED="\e[31m"
+GREEN="\e[32m"
+RESETSTYLE="\e[0m"
+BOLD="\e[1m"
+
+function initTest() {
+    echo -e "${BOLD}\n🏃 ${1}${RESETSTYLE}"
```
nickboldt, line 26: emojis sometimes break stuff... is this needed?

build/dockerfiles/test_entrypoint.sh
```diff
@@ -0,0 +1,350 @@
...
+    else
+        echo -e "${RED}✘${RESETSTYLE} Test failed!"
+        echo "Result:"
+        cat "${file}"
+        echo "Expected:"
+        echo "${expected_metayaml}"
+        exit 1
+    fi
+}
+echo -e "${BOLD}\n🏃🏃🏃 Running tests for entrypoint.sh: ${BASH_SOURCE[0]}${RESETSTYLE}"
```
nickboldt, line 49: sometimes when you copy console code into editors (like JIRA) it breaks when trying to save the non-ascii code to the database. wondering what the value is here of a running icon repeated 3 times when the next word is also "Running" :)

nickboldt (approved):
minor nitpick, otherwise LGTM (skimmed only, an expert should approve it too)

build/dockerfiles/test_entrypoint.sh
sunix: for test, visually it looks better when reading the logs in CI

sunix: For instance: 
Selection_347

build/dockerfiles/test_entrypoint.sh
sunix: i would say it is a bug in JIRA. or screenshot. The goal of this emojis is to visually helps someone reading CI logs and quickly figure out what is going on and why it is failing if it fails.

sunix: @nickboldt I made the emoji optional (could be overridden with env variable)

nickboldt: cool, might help when using consoles on older terminal emulators. 

'twas just a nitpick (and yeah, it's a JIRA bug/sql limitation, I agree).

build/dockerfiles/test_entrypoint.sh
nickboldt: :+1:

ericwill (approved):
Works as expected on my side :+1:

PR #958, opened 2021-05-17
feat(python-jupyter): Setting up venv in a dedicated volume

sunix:
Signed-off-by: Sun Seng David TAN <sutan@redhat.com>

### What does this PR do?
Sidecar part, using the same venv folder for:
- jupyter notebook from the python plugin
- projects with dependencies in `requirements.txt`

https://youtu.be/FPZa5jc8sYE

https://github.com/eclipse/che/issues/16818

### How to test this PR?
will come with the plugin part (in a second PR once this is merged) but could be tested with:
```yaml
apiVersion: 1.0.0
metadata:
  generateName: python-jupyter-
projects:
  - name: jupyter-hello-world1
    source:
      startPoint: master
      location: 'https://github.com/chasbecker/TextAnalysis.git'
      type: git
components:
  - id: ms-python/python/latest
    registryUrl: 'https://sunix-che-plugin-registry-dev-i1viq.surge.sh/v3'
    type: chePlugin
  - mountSources: true
    memoryLimit: 512Mi
    type: dockerimage
    volumes:
      - name: venv
        containerPath: /home/user/.venv
    image: 'quay.io/eclipse/che-python-3.8:7.29.0'
    alias: python
commands:
  - name: set up venv with requirements
    actions:
      - workdir: /home/user
        type: exec
        command: python -m venv .venv && . .venv/bin/activate && python -m pip install -r /projects/jupyter-hello-world1/requirements.txt
        component: python
```
and the flow:
1. open `AzureTextAnalyticsRestAPI.ipynb`
2. wait for the jupyter notebook to be loaded
3. try to run the first lines with imports
4. it should fail with errors that could not find the module `requests`
5. run the task `set up venv with requirements`
6. try to run again the first lines with imports
7. it should now fail: cannot find `pandas`
8. add to the file `requirements.txt` 
   ```
   pandas==1.2.4
   ```
9. run the task `set up venv with requirements`
10. try to run again the first lines with imports
11. it should now works, even the next part of the notebook

### PR Checklist

### Reviewers

### DIFF
--- a/sidecars/python/etc/entrypoint.sh
+++ b/sidecars/python/etc/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Copyright (c) 2018-2020 Red Hat, Inc.
 # This program and the accompanying materials are made
@@ -20,7 +20,7 @@ GROUP_ID=$(id -g)
 export GROUP_ID
 
 if ! whoami >/dev/null 2>&1; then
-    echo "${USER_NAME:-user}:x:${USER_ID}:0:${USER_NAME:-user} user:${HOME}:/bin/sh" >> /etc/passwd
+    echo "${USER_NAME:-user}:x:${USER_ID}:0:${USER_NAME:-user} user:${HOME}:/bin/bash" >> /etc/passwd
 fi
 
 # Grant access to projects volume in case of non root user with sudo rights
@@ -28,4 +28,13 @@ if [ "${USER_ID}" -ne 0 ] && command -v sudo >/dev/null 2>&1 && sudo -n true > /
     sudo chown "${USER_ID}:${GROUP_ID}" /projects
 fi
 
+# Setup .venv in the 'venv' volume that should be mounted in HOME/.venv
+mkdir -p "${HOME}"/.venv
+if [ ! -f "${HOME}"/.venv/bin/activate ]; then
+  echo "${HOME}"/.venv is empty, moving files from "${HOME}"/.venv-tmp/
+  mv "${HOME}"/.venv-tmp/* "${HOME}"/.venv
+fi
+
+# shellcheck source=/dev/null
+source "${HOME}"/.venv/bin/activate
 exec "$@"
--- a/sidecars/python/Dockerfile
+++ b/sidecars/python/Dockerfile
@@ -12,13 +12,6 @@ FROM python:3.8.6-slim
 
 ENV HOME=/home/theia
 
-RUN mkdir /projects ${HOME} && \
-    # Change permissions to let any arbitrary user
-    for f in "${HOME}" "/etc/passwd" "/projects"; do \
-      echo "Changing permissions on ${f}" && chgrp -R 0 ${f} && \
-      chmod -R g+rwX ${f}; \
-    done
-
 RUN apt-get update && \
     apt-get install exuberant-ctags && \
     apt-get install wget -y && \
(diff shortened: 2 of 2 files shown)


sidecars/python/etc/entrypoint.sh
```diff
@@ -28,4 +28,13 @@ if [ "${USER_ID}" -ne 0 ] && command -v sudo >/dev/null 2>&1 && sudo -n true > /
     sudo chown "${USER_ID}:${GROUP_ID}" /projects
 fi
 
+# Setup .venv in the 'venv' volume that should be mounted in HOME/.venv
+cd "${HOME}"
+python -m venv .venv
+# shellcheck disable=SC1091
+. .venv/bin/activate
+# Setup jupyter dependencies in this venv
+pip install -U ipykernel jupyter
```
ericwill, line 37: Does this need to be here? Or could it be in the project's `requirements.txt` file?

sidecars/python/etc/entrypoint.sh
```diff
@@ -28,4 +28,13 @@ if [ "${USER_ID}" -ne 0 ] && command -v sudo >/dev/null 2>&1 && sudo -n true > /
     sudo chown "${USER_ID}:${GROUP_ID}" /projects
 fi
 
+# Setup .venv in the 'venv' volume that should be mounted in HOME/.venv
+cd "${HOME}"
+python -m venv .venv
+# shellcheck disable=SC1091
+. .venv/bin/activate
```
ericwill, line 35: Convention is usually to use `source`, we should probably install bash in the sidecar container for this reason

sunix: any sources of this convention?

sidecars/python/etc/entrypoint.sh
sunix: Looks like projects using jupyter notebook are not adding these dependencies in `requirements.txt`. As the python extension provide the jupyter notebook feature, we are installing these dependencies so it works out of the box as well as providing a smooth user experience.

sidecars/python/etc/entrypoint.sh
ericwill: Yes, here: https://docs.python.org/3/tutorial/venv.html#creating-virtual-environments

sidecars/python/etc/entrypoint.sh
ericwill: If this is the case maybe we should install jupyter in the Dockerfile for the python sidecar, instead of in the entrypoint?

sidecars/python/etc/entrypoint.sh
sunix: ok let's install bash

sidecars/python/etc/entrypoint.sh
sunix: it doesn't work because the dependencies installed won't be available in the venv. It failed when I tried. It is done in the entrypoint so we can perform the venv setup in the volume (that is available at runtime).

ericwill: Fine by me then :+1:

benoitf: I think doing post-install step in the entrypoint.sh is not a viable solution

sunix: @benoitf why not? what is your proposal?  having that we may have additional dependencies coming in the same venv from a projects

sidecars/python/etc/entrypoint.sh
sunix: actually bash is already installed!

sidecars/python/etc/entrypoint.sh
benoitf: I mean, the goal is to have the containers starting as fast as possible.

doing that in the entrypoint is delaying the startup of the workspace and it can even fail to start the workspace if we're unable to install the dependencies, etc.


For example in maven case, it's vscode-java that pull maven dependencies but after the workspace is started, not in the entrypoint.

ericwill: It also just occurred to me that we need to think about restricted network/self-signed certificate environments: https://www.eclipse.org/che/docs/che-7/end-user-guide/using-python-artifact-repositories/

I think the current approach will break that

sunix: a container is considered 'started' before the entrypoint is executed. otherwise with `sleep infinity` we would never have any container `started`.

sunix: @ericwill it should work, but it would be worth testing

sunix: I can see how I can make it more robust, and not failing if we have network issues.

sunix: ok i will try https://stackoverflow.com/questions/7438681/how-to-duplicate-virtualenv

sunix: or just do the venv setup in the dockerfile and make a copy in the entrypoint

benoitf: yes, being able to just copy pre-pulled environment would be better option

benoitf (approved):
Screenshot 2021-05-25 at 11 41 28


If you could merge the RUN instructions to reduce the number of layers

One RUN instead of 3

sunix:
@benoitf ... for some reasons when i tried to merge, the gh actions are broken ... it works well locally with podman but it is hard for me to fix it as i cannot reproduce. OK to merge like that ?

PR #960, opened 2021-05-19
[base] Update from rhscl/httpd-24-rhel7:2.4-133.1608869526 to rhscl/httpd-24-rhel7:2.4-143.1619800879

nickboldt:
Change-Id: Ib7a242515e45e5ec6773fc9b94be2a09e940c4f5

[base] Update from rhscl/httpd-24-rhel7:2.4-133.1608869526 to rhscl/httpd-24-rhel7:2.4-143.1619800879

Change-Id: Ib7a242515e45e5ec6773fc9b94be2a09e940c4f5

### DIFF
--- a/build/dockerfiles/rhel.Dockerfile
+++ b/build/dockerfiles/rhel.Dockerfile
@@ -13,7 +13,7 @@
 # Build registry
 # UPSTREAM: use RHEL7/RHSCL/httpd image so we're not required to authenticate with registry.redhat.io
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/rhscl/httpd-24-rhel7
-FROM registry.access.redhat.com/rhscl/httpd-24-rhel7:2.4-133.1608869526
+FROM registry.access.redhat.com/rhscl/httpd-24-rhel7:2.4-143.1619800879
 
 # DOWNSTREAM: use RHEL8/httpd
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/rhel8/httpd-24


build/dockerfiles/rhel.Dockerfile
```diff
@@ -13,11 +13,11 @@
 # Build registry
 # UPSTREAM: use RHEL7/RHSCL/httpd image so we're not required to authenticate with registry.redhat.io
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/rhscl/httpd-24-rhel7
-FROM registry.access.redhat.com/rhscl/httpd-24-rhel7:2.4-133.1608869526
+FROM registry.access.redhat.com/rhscl/httpd-24-rhel7:2.4-143.1619800879
```
benoitf, line 16: should we not switch instead to `ubi8/httpd-24` ?

nickboldt: > # UPSTREAM: use RHEL7/RHSCL/httpd image so we're not required to authenticate with registry.redhat.io

nickboldt: I believe this is one of those situations where the image that's available to the public is on one URL/registry and the equivalent image internally that we need for Brew is different. So we could maybe use ubi8/httpd upstream and rhel8/httpd downstream, or httpd-rhel7 up and rhel8/httpd down. 

Either way, needs to be different... and this works. Since we don't BUILD this rhel.dockerfile upstream in any tests, doesn't really matter what the hypothetical base image is. Or... do we have users that use `build.sh --rhel` here to use this dockerfile instead of the alpine one?

benoitf: > UPSTREAM: use RHEL7/RHSCL/httpd image so we're not required to authenticate with registry.redhat.io

I know that but ubi8/httpd-24 is not requiring authentication as well

benoitf: > Either way, needs to be different... and this works. Since we don't BUILD this rhel.dockerfile upstream in any tests, doesn't really matter what the hypothetical base image is. Or... do we have users that use build.sh --rhel here to use this dockerfile instead of the alpine one?


we do run happy path tests on both alpine and rhel images (and we check that image is building as well)

ericwill: If we can use the ubi8 image upstream without authentication then I'm +1 to that

nickboldt: ubi8/httpd is new as of UBI 8.4, so if that works w/o auth, happy to switch to that

ericwill (approved):

nickboldt:
cherrypicked in 7.30.x as https://github.com/eclipse-che/che-plugin-registry/commit/a8cd9b3f417c23cd5bd277acf5bbdb7a8adbcf3d

PR #1014, opened 2021-07-30
fix: CRW-2118: Contributing security updates for httpd

ArvinB:
This will deny access to security sensitive .ht files and increase the default encryption level for SSL.

https://issues.redhat.com/browse/CRW-2118

### PR Checklist

- [X] [The Eclipse Contributor Agreement is valid](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-eclipse-contributor-agreement-is-valid)
- [X] [Code produced is complete](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-produced-is-complete)
- [X] [Code builds without errors](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-builds-without-errors)
- [X] [Tests are covering the bugfix](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#tests-are-covering-the-bugfix)
- [X] [The repository devfile is up to date and works](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-repository-devfile-is-up-to-date-and-works)
- [X] [Sections `What issues does this PR fix or reference` and `How to test this PR` completed](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#sections-what-issues-does-this-pr-fix-or-reference-and-how-to-test-this-pr-completed)
- [X] [Relevant user documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [X] [Relevant contributing documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [X] [CI/CD changes implemented, documented and communicated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#cicd-changes-implemented-documented-and-communicated)

### Reviewers

We use these settings downstream in IBM Wazi Developer for Workspaces.

### DIFF
--- a/build/dockerfiles/rhel.Dockerfile
+++ b/build/dockerfiles/rhel.Dockerfile
@@ -21,10 +21,6 @@ RUN chmod +x /usr/share/container-scripts/httpd/pre-init/40-ssl-certs.sh && \
 RUN yum update -y systemd && yum clean all && rm -rf /var/cache/yum && \
     echo "Installed Packages" && rpm -qa | sort -V && echo "End Of Installed Packages"
 
-RUN echo "<FilesMatch "\""^\\.ht"\"">" >> /etc/httpd/conf/httpd.conf && \
-    echo "Require all denied" >> /etc/httpd/conf/httpd.conf && \
-    echo "</FilesMatch>" >> /etc/httpd/conf/httpd.conf
-
 RUN sed -i /etc/httpd/conf.d/ssl.conf \
     -e "s,SSLProtocol all -SSLv2,SSLProtocol all -SSLv3," \
     -e "s,SSLCipherSuite HIGH:MEDIUM:!aNULL:!MD5,SSLCipherSuite HIGH:!aNULL:!MD5,"
--- a/build/dockerfiles/rhel.Dockerfile
+++ b/build/dockerfiles/rhel.Dockerfile
@@ -8,6 +8,7 @@
 #
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
+#   IBM Corporation - implementation
 #
 
 # Build registry
@@ -20,7 +21,14 @@ RUN chmod +x /usr/share/container-scripts/httpd/pre-init/40-ssl-certs.sh && \
 RUN yum update -y systemd && yum clean all && rm -rf /var/cache/yum && \
     echo "Installed Packages" && rpm -qa | sort -V && echo "End Of Installed Packages"
 
-# BEGIN these steps might not be required
+RUN echo "<FilesMatch "\""^\\.ht"\"">" >> /etc/httpd/conf/httpd.conf && \
+    echo "Require all denied" >> /etc/httpd/conf/httpd.conf && \
+    echo "</FilesMatch>" >> /etc/httpd/conf/httpd.conf
+
+RUN sed -i /etc/httpd/conf.d/ssl.conf \
+    -e "s,SSLProtocol all -SSLv2,SSLProtocol all -SSLv3," \
+    -e "s,SSLCipherSuite HIGH:MEDIUM:!aNULL:!MD5,SSLCipherSuite HIGH:!aNULL:!MD5,"
+
 RUN sed -i /etc/httpd/conf/httpd.conf \
     -e "s,Listen 80,Listen 8080," \
     -e "s,logs/error_log,/dev/stderr," \
@@ -28,7 +36,6 @@ RUN sed -i /etc/httpd/conf/httpd.conf \
     -e "s,AllowOverride None,AllowOverride All," && \
     chmod a+rwX /etc/httpd/conf /run/httpd /etc/httpd/logs/
 STOPSIGNAL SIGWINCH
-# END these steps might not be required
 
 WORKDIR /var/www/html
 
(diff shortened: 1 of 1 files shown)


ArvinB:
@nickboldt @ericwill FYI

build/dockerfiles/rhel.Dockerfile
```diff
@@ -20,15 +21,21 @@ RUN chmod +x /usr/share/container-scripts/httpd/pre-init/40-ssl-certs.sh && \
 RUN yum update -y systemd && yum clean all && rm -rf /var/cache/yum && \
     echo "Installed Packages" && rpm -qa | sort -V && echo "End Of Installed Packages"
 
-# BEGIN these steps might not be required
+RUN echo "<FilesMatch "\""^\\.ht"\"">" >> /etc/httpd/conf/httpd.conf && \
+    echo "Require all denied" >> /etc/httpd/conf/httpd.conf && \
+    echo "</FilesMatch>" >> /etc/httpd/conf/httpd.conf
+
+RUN sed -i /etc/httpd/conf.d/ssl.conf \
```
benoitf, line 28: hello, could you describe what it does fix ?
AFAIK ssl engine is not enabled (it is listening on http port)

nickboldt (approved):

build/dockerfiles/rhel.Dockerfile
```diff
@@ -20,15 +21,21 @@ RUN chmod +x /usr/share/container-scripts/httpd/pre-init/40-ssl-certs.sh && \
 RUN yum update -y systemd && yum clean all && rm -rf /var/cache/yum && \
     echo "Installed Packages" && rpm -qa | sort -V && echo "End Of Installed Packages"
 
-# BEGIN these steps might not be required
+RUN echo "<FilesMatch "\""^\\.ht"\"">" >> /etc/httpd/conf/httpd.conf && \
```
benoitf, line 24: If I look at the current file
```
$ docker run --rm -it --entrypoint=sh registry.access.redhat.com/ubi8/httpd-24:1-143 -c "cat /etc/httpd/conf/httpd.conf"
```

 it contains

```#
# The following lines prevent .htaccess and .htpasswd files from being
# viewed by Web clients.
#
<Files ".ht*">
    Require all denied
</Files>
```

benoitf (changes requested):
please see current comments

build/dockerfiles/rhel.Dockerfile
ArvinB: @benoitf Thank you. I will remove this section.

ArvinB: Removed

build/dockerfiles/rhel.Dockerfile
ArvinB: Downstream SSL is used and IBM security flagged that the encryption strength could be better. Thus the entries for the Cipher Suite and Protocol.

benoitf: @ArvinB from what I see there, ssl is started, but the 8080 port is used and not the 8043
so either the issue is to not start ssl port or then it needs to use 8043 port in deployments (which is not the case in operator or in helm charts)

ArvinB: @benoitf Yes, I see your point and also see port 8080 is used. Even if the ssl port (8043) is not started. We have guidance to still setup various security provisions within the config for any container image that IBM publishes. That way if the service is turned on and used, then it is already secured.

So I am requesting please if we can keep this config upstream here, because it will make it's way downstream into an IBM published container image.

Reference from OpenShift:
image

benoitf: I would say it's misleading to say it's solving security issues that we don't have as we never use that part.
Why are we even copy ssl config file if it's never used.

In case you really want to keep such useless configuration, I would recommend to avoid any sed expression that can break at any time silently. For example SSLProtocol property entry can be updated and you think sed is replacing but it may not.

So probably sed expression need to be more robust if you want to keep it and check that it's correctly applied.

ArvinB: Hi @benoitf 

I can agree with that, it is misleading to say that we are solving a security issue. As the PR title says, it's more accurate to simply say that security updates are being contributed into the httpd configs.

We do want to keep the config. So we can put something else in place instead of using a `sed` expression. I am open to an alternative recommendation? However, using `sed` is pretty common and well used all the time. Is the concern that there isn't a check in place to ensure the settings have been applied properly?

benoitf: @ArvinB the sed expression you used assume to replace a property + a default value so a safer sed expression would be to replace the value whatever is the default value is

And yes it would be great to ensure the property has been applied

ArvinB: I will do this at the end of next week. Thank you for the quick response.

ArvinB: @benoitf 
Sorry for the delay, work and life has been nutz!
I just pushed changes based on our convo.

benoitf (approved):

PR #1810, opened 2023-10-10
fix: use https protocol in the 'che-incubator/che-code/insiders' editor

ibuziuk:
### What does this PR do?

use https protocol in the 'che-incubator/che-code/insiders' editor

Note that it only changes protocol for `che-incubator/che-code/insiders`, not sure if we should update it for the rest of the editors (some of which look obsolete now and should probably be removed).

Screenshot 2023-10-10 at 16 29 18

https://issues.redhat.com/browse/CRW-4869

### How to test this PR?
Tested with custom che-editor.yaml - https://github.com/ibuziuk/quarkus-api-example/blob/main/.che/che-editor.yaml

### PR Checklist

- [ ] [The Eclipse Contributor Agreement is valid](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-eclipse-contributor-agreement-is-valid)
- [ ] [Code produced is complete](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-produced-is-complete)
- [ ] [Code builds without errors](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-builds-without-errors)
- [ ] [Sections `What issues does this PR fix or reference` and `How to test this PR` completed](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#sections-what-issues-does-this-pr-fix-or-reference-and-how-to-test-this-pr-completed)

### Reviewers

### DIFF
--- a/che-editors.yaml
+++ b/che-editors.yaml
@@ -295,25 +295,25 @@ editors:
               targetPort: 3100
               exposure: public
               secure: true
-              protocol: http
+              protocol: https
             - name: code-redirect-1
               targetPort: 13131
               exposure: public
-              protocol: http
+              protocol: https
               attributes:
                 discoverable: false
                 urlRewriteSupported: false
             - name: code-redirect-2
               targetPort: 13132
               exposure: public
-              protocol: http
+              protocol: https
               attributes:
                 discoverable: false
                 urlRewriteSupported: false
             - name: code-redirect-3
               targetPort: 13133
               exposure: public
-              protocol: http
+              protocol: https
               attributes:
                 discoverable: false
                 urlRewriteSupported: false
--- a/che-editors.yaml
+++ b/che-editors.yaml
@@ -208,25 +208,25 @@ editors:
               targetPort: 3100
               exposure: public
               secure: true
-              protocol: http
+              protocol: https
             - name: code-redirect-1
               targetPort: 13131
               exposure: public
-              protocol: http
+              protocol: https
               attributes:
                 discoverable: false
                 urlRewriteSupported: false
             - name: code-redirect-2
               targetPort: 13132
               exposure: public
-              protocol: http
+              protocol: https
               attributes:
                 discoverable: false
                 urlRewriteSupported: false
             - name: code-redirect-3
               targetPort: 13133
               exposure: public
-              protocol: http
+              protocol: https
               attributes:
                 discoverable: false
                 urlRewriteSupported: false
--- a/tools/build/src/devfile/meta-yaml-to-devfile-yaml.ts
+++ b/tools/build/src/devfile/meta-yaml-to-devfile-yaml.ts
@@ -131,13 +131,14 @@ export class MetaYamlToDevfileYaml {
           devfileEndpoint.attributes['type'] = 'main';
         }
 
-        // if it's secured, remove secure option for now but add extra s on the protocol
+        // if it's secured, remove 'secure' option for now but change protocol to secure (https or wss)
         if (devfileEndpoint.attributes && devfileEndpoint.attributes.secure === true) {
           devfileEndpoint.secure = false;
           delete devfileEndpoint.attributes['secure'];
-          // add extra s
-          if (devfileEndpoint.attributes.protocol) {
-            devfileEndpoint.attributes.protocol = `${devfileEndpoint.attributes.protocol}s`;
+          // switch to secure protocol (https or wss)
+          const protocol = devfileEndpoint.attributes.protocol;
+          if (protocol && (protocol === 'http' || protocol === 'ws')) {
+            devfileEndpoint.attributes.protocol = `${protocol}s`;
           }
         }
 
--- a/tools/build/tests/_data/meta/che-code-meta.yaml
+++ b/tools/build/tests/_data/meta/che-code-meta.yaml
@@ -19,7 +19,7 @@ spec:
         cookiesAuthEnabled: true
         discoverable: false
         urlRewriteSupported: true
-        protocol: http
+        protocol: https
         secure: true
       targetPort: 3100
       public: true
@@ -28,21 +28,21 @@ spec:
       attributes:
         discoverable: false
         urlRewriteSupported: false
-        protocol: http
+        protocol: https
       public: true
     - name: code-redirect-2
       targetPort: 13132
       attributes:
         discoverable: false
         urlRewriteSupported: false
-        protocol: http
+        protocol: https
       public: true
     - name: code-redirect-3
       targetPort: 13133
       attributes:
         discoverable: false
         urlRewriteSupported: false
-        protocol: http
+        protocol: https
       public: true
   containers:
     - image: quay.io/devfile/universal-developer-image:latest
--- a/tools/build/tests/common/endpoints-helper.spec.ts
+++ b/tools/build/tests/common/endpoints-helper.spec.ts
@@ -49,6 +49,26 @@ describe('Test EndpointsHelper', () => {
     expect(endpoint.secure).toBeUndefined();
   });
 
+  test('https', async () => {
+    const endpointYaml: CommonEndpointYaml = {
+      name: 'endpoint-name',
+      path: 'endpoint-path',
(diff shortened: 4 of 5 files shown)


l0rd:
I would set `https` to `che-incubator/che-code/7.75.0` endpoints too (you only applied to `che-incubator/che-code/insiders`).

ibuziuk:
@l0rd I can do that, but I thought that 7.75.0 has been released and we should not update it afterwards

svor:
we need to change `che-incubator/che-code/7.75.0` as well it means like a latest version. 
You can check the generated registry of this PR: https://pr-check-1810-che-plugin-registry.surge.sh/v3/plugins/che-incubator/che-code/
It provides 2 che-code editors: latest and isiders but latest doesn't have http --> https changes 

probably we need the same changes in downstream as well: https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/dependencies/che-plugin-registry/che-editors.yaml#L166

ibuziuk:
thanks for the clarification, updated

ibuziuk:
@svor looks like smth. weird is happening with the `che-code` endpoint resulting in `httpss` protocol. Looks like the script explicitly adds extra `s` enforcing the https, but not checking if `https` is already set.  Not sure if we should update the script, or just do not update `che-code` endpoint, which would look weird imo


Screenshot 2023-10-10 at 16 15 53

tools/build/src/devfile/meta-yaml-to-devfile-yaml.ts
```diff
@@ -131,13 +131,14 @@ export class MetaYamlToDevfileYaml {
           devfileEndpoint.attributes['type'] = 'main';
         }
 
-        // if it's secured, remove secure option for now but add extra s on the protocol
+        // if it's secured, remove 'secure' option for now but change protocol to secure (https or wss)
         if (devfileEndpoint.attributes && devfileEndpoint.attributes.secure === true) {
           devfileEndpoint.secure = false;
```
svor, line 136: I don't understand why we change **secure** to false when it was set to true.
I think we shouldn't change the value.

ibuziuk: @l0rd do you happen to remember why it was done that way? I recall that `secure` attribute might be in general not supported by DWO

svor (approved):
Workspace was started successfully with current editor's configuration
screenshot-che-dogfooding apps che-dev x6e0 p1 openshiftapps com-2023 10 10-22_58_01

tools/build/src/devfile/meta-yaml-to-devfile-yaml.ts
l0rd: No, I suspect that `secured` is currently ignored (it was used by che-server when it was creating routes) and to have a secured endpoint (that require authentication) you need to set the `supportURLRewrite` to true.

amisevsk: The Devfile API docs give

> Describes whether the endpoint should be secured and protected by some authentication process. This requires a protocol of https or wss.

but I think (haven't verified) that this field is either ignored in Che or is interpreted as "use https instead of http" (which isn't really appropriate, API-wise). I think the "most correct" solution would be something like

* If `secured` and `supportURLRewrite`, use gateway URL
* If either is false, use plain route
* If `secured` but not `supportURLRewrite`, fail workspace with error as we can't put auth on a plain route

che-editors.yaml
```diff
@@ -208,25 +208,25 @@ editors:
...
-              protocol: http
+              protocol: https
             - name: code-redirect-1
               targetPort: 13131
               exposure: public
-              protocol: http
+              protocol: https
               attributes:
                 discoverable: false
                 urlRewriteSupported: false
```
amisevsk, line 218: I haven't had time to verify this directly and so I'm not sure it's an _actual_ problem, but does this cause issues when running on Kubernetes? Since these endpoints have `urlRewriteSupported: false`, I believe Che will support them via plain ingresses, and I'm not sure if Che is able to configure TLS there.

tools/build/src/devfile/meta-yaml-to-devfile-yaml.ts
ibuziuk: @l0rd @amisevsk I would probably leave it as-is for now (just fixing `httpss / wsss`cases), and I think we need a separate issue for supporting `secure` attribute properly  and not do any weird patching on plugin-registry end. Please, approve if this makes sense

che-editors.yaml
ibuziuk: yeah, good, I need to check this on minikube

ibuziuk: tried to verify on minikube but it looks like Che is not supported on this arch in general:

```
 ❯ Eclipse Che Operator pod bootstrap
      ✔ Scheduling...[OK]
      ✖ Downloading images
        → Failed to download image, reason: ErrImagePull, message: rpc error: co
…
        Starting
      Create ValidatingWebhookConfiguration org.eclipse.che
      Create MutatingWebhookConfiguration org.eclipse.che
      Create CheCluster Custom Resource
Error: Command server:deploy failed with the error: Failed to download image, reason: ErrImagePull, message: rpc error: code = Unknown desc = no matching manifest for linux/arm64/v8 in the manifest list entries. See details: /Users/ibuziuk/Library/Caches/chectl/error.log. Eclipse Che logs: /var/folders/87/t16qbhjd6sn7cz51_g716sh40000gn/T/chectl-logs/1697121281363.
    at newError (/Users/ibuziuk/chectl/lib/utils/utls.js:39:19)
    at wrapCommandError (/Users/ibuziuk/chectl/lib/utils/command-utils.js:53:32)
    at Deploy.<anonymous> (/Users/ibuziuk/chectl/lib/commands/server/deploy.js:82:65)
    at Generator.throw (<anonymous>)
    at rejected (/Users/ibuziuk/chectl/node_modules/tslib/tslib.js:165:69)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
Cause: Error: Failed to download image, reason: ErrImagePull, message: rpc error: code = Unknown desc = no matching manifest for linux/arm64/v8 in the manifest list entries.
    at /Users/ibuziuk/chectl/lib/tasks/pod-tasks.js:149:39
    at Generator.next (<anonymous>)
    at fulfilled (/Users/ibuziuk/chectl/node_modules/tslib/tslib.js:164:62)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

However, Smoke Test PR check should cover the minikube verification. Adding @dmytro-ndp for PR approval

dmytro-ndp: I had gotten Network Error when tried to start workspace of Eclipse Che Next on minikube which is using external plugin registry with URL https://pr-check-1810-che-plugin-registry.surge.sh/v3 given by @ibuziuk :
Screenshot from 2023-10-17 13-30-08

dmytro-ndp: BTW: the error `Failed to fetch a devfile from URL: https://pr-check-1810-che-plugin-registry.surge.sh/v3/plugins/che-incubator/che-code/latest/devfile.yaml, reason: Error: Network Error` has reproduced on OpenShift 4.10 as well:
Screenshot from 2023-10-17 16-15-29

ibuziuk: @olexii4 @akurinnoy folks, it looks like in general it is not possible to reference external plugin registry from the CR since we would hit CORS on dashboard end, right?

ibuziuk: It looks like CORS requests are disabled on PR checks like https://pr-check-1810-che-plugin-registry.surge.sh/v3 
@svor do you happen to know if it is expected?

svor: I've just tested it on minikube (the workspace has been started, all plugins have been installed):
screenshot-192 168 49 2 nip io-2023 10 30-17_03_01
screenshot-192 168 49 2 nip io-2023 10 30-17_02_25

